PR #2, opened 2015-01-20
Add Jenkins Recipe for chef-backup Job

bramwelt:
Ensures the `/var/chef-backup-for-rdiff` directory exists and is owned by jenkins, and the `knife-backup` plugin is installed.


dspt:
action :install will not specify a version for the gem, so I think it's better to use action :upgrade to ensure the latest version, or version "$version" action :install to pin it at a certain version
http://docs.chef.io/resource_chef_gem.html

bramwelt:
Pinning it to `0.0.10`.

jordane:
@bramwelt A comment explaining which job this is for would be good. Or maybe even adding the job to the recipe.

ramereth:
Tests?

bramwelt:
Tests added along with documentation for the recipe.

.kitchen.cloud.yml
```diff
@@ -15,8 +15,8 @@ driver_config:
 
 provisioner:
   name: chef_solo
-  data_bags_path: test/integration/default/data_bags
-  encrypted_data_bag_secret_key_path: test/integration/default/encrypted_data_bag_secret
+  #data_bags_path: test/integration/default/data_bags
```
ramereth: Why is this commented out?

recipes/chef-backup.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+chef_gem 'knife-backup' do
+    action :install
```
ramereth: Shouldn't this be just two space indents?

.kitchen.cloud.yml
bramwelt: These cause `kitchen` to fail if they don't exist. Since the cookbook doesn't use data_bags, I have removed these lines.

test/integration/chef_backup/serverspec/chef_backup_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'serverspec'
+
+set :backend, :exec
+
+RSpec.configure do |config|
+    config.before(:all) do
```
ramereth: 4 space indent?

bramwelt:
Indentation and `.kitchen.cloud.yml` fixed.

test/integration/chef_backup/serverspec/spec_helper.rb
```diff
@@ -0,0 +1,9 @@
+require 'serverspec'
+
+set :backend, :exec
+
+RSpec.configure do |config|
+  config.before(:all) do
+    config.path = '/opt/chef/embedded/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
```
ramereth: Line is over 80char.

ramereth:
Rubocop also complains about the name of the file:

```
recipes/chef-backup.rb:1:1: C: Use snake_case for source file names.
```

I believe they want it renamed to ``chef_backup.rb``. @jordane thoughts on doing that to keep rubocop happy?

jordane:
> I believe they want it renamed to chef_backup.rb. @jordane thoughts on doing that to keep rubocop happy?

Yes, this is a standard ruby thing. We should follow it.

armiller:
:+1:

ramereth:
@bramwelt mind bumping to CentOS 6.6?

armiller:
@osuosl-cookbooks/chefs I updated the kitchen configs and also added rubocop. I re-ran test kitchen with 6.6 and everything passed.

:+1:

ramereth:
Mind merging the two test files into one?

armiller:
ok. good to go.

ramereth:
:+1:

test/integration/chef_backup/serverspec/chef_backup_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+end
+
+describe package('knife-backup') do
+  it { should be_installed.by('gem').with_version('0.0.10') }
+end
+
+describe file('/var/chef-backup-for-rdiff') do
+  it { should be_directory }
+  it { should be_owned_by 'centos' }
+  it { should be_grouped_into 'centos' }
```
rettigs: Why are the user and group hardcoded when they're set by the attributes `node['jenkins']['master']['{user,group}']` in `recipes/chef_backup.rb`?

jordane: Serverspec has no concept of attributes.

bramwelt:
@armiller Thanks so much for picking this up for me!

PR #3, opened 2015-02-26
Cookbook Uploader Magic

jordane:
This will create jenkins jobs that automatically upload cookbooks based on tags to the chef server. It will create said jobs for any repo in `osuosl-cookbooks`.

Edit: https://github.com/osuosl/chef-repo/pull/327 must be merged before merging this!


ramereth:
@jordane is there anyway to test this with serverspec?

jordane:
@ramereth Yes. It creates a bunch of xml files, and we can at least test for those. It'd be really hard to test the jenkins jobs themselves though..

templates/default/cookbook-uploader.config.xml.erb
```diff
@@ -0,0 +1,69 @@
...
+      <ignorePostCommitHooks>false</ignorePostCommitHooks>
+    </hudson.triggers.SCMTrigger>
+  </triggers>
+  <concurrentBuild>false</concurrentBuild>
+  <customWorkspace>/var/lib/jenkins/jenkins-data/jobs/<%= @org %>/<%= @repo %></customWorkspace>
+  <builders>
+    <hudson.tasks.Shell>
+      <command>export KNIFE_USER=&apos;jenkins&apos;
+pwd
+knife cookbook upload -o ../ $(grep ^name metadata.rb| awk &apos;{print $2;}&apos; | tr -d \&apos;)</command>
```
armiller: * What is `&apos`? Is that for whitespace? 
* Are we not freezing the cookbook on upload?

jordane: @armiller that gets rendered as `'`.

The lack of `--freeze` was an oversight.

bramwelt:
@jordane Shouldn't this be testable when using chef_zero? I'd guess the cookbooks that get 'uploaded' in this case will get stored on the filesystem for chef_zero.

.kitchen.cloud.yml
```diff
@@ -17,17 +17,23 @@ provisioner:
...
     run_list:
       - recipe[osl-jenkins::default]
     attributes:
+  - name: cookbook-uploader
+    run_list:
+      - recipe[java]
+      - recipe[jenkins::master]
+      - recipe[osl-jenkins::cookbook-uploader]
+    data_bags_path: ../../data_bags/
+    encrypted_data_bag_secret_key_path: ../../../../.chef/encrypted_data_bag_secret
```
bramwelt: It's best to create a new `data_bags` and `encrypted_data_bag_secret` for testing purposes under `test/`

jordane:
@bramwelt One of the major problems with testing this with Chef Zero is that Chef Zero doesn't run all the time, it only runs when the client spins it up. We'd have to write a shim to spin it up before any of the jobs ran.

recipes/cookbook-uploader.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+secrets = Chef::EncryptedDataBagItem.load(node['osl-jenkins']['databag'], 'secrets')
+repos = collect_github_repositories(secrets, 'osuosl-cookbooks')
```
bramwelt: As `secrets` here is the databag item: `node['osl-jenkins']['databag']:secrets` I'm thinking it would contain more than a github token (including things like `id` and `name`). Shouldn't this be `secrets['github-token']` or something similar?

recipes/cookbook-uploader.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+  template xml do
+    source 'cookbook-uploader.config.xml.erb'
+  end
+  jenkins_job repo do
+    config xml
+    action [:create, :enable]
+  end
+end
+
```
bramwelt: Extra newline.

recipes/cookbook-uploader.rb
jordane: Nope. Go look closer at what `collect_github_repositories` does.

bramwelt: Ah, I see now.

.kitchen.cloud.yml
jordane: Yeah, the reason this was done originally was so that an oauth token wasn't committed in this repo, but it turns out that octokit doesn't fail if you pass it an invalid token, it just won't log you in. so we can still test against pub repos.

templates/default/cookbook-uploader.config.xml.erb
```diff
@@ -0,0 +1,69 @@
...
+  <description></description>
+  <keepDependencies>false</keepDependencies>
+  <properties/>
+  <scm class="hudson.plugins.git.GitSCM" plugin="git@1.4.0">
+    <configVersion>2</configVersion>
+    <userRemoteConfigs>
+      <hudson.plugins.git.UserRemoteConfig>
+        <name>origin</name>
+        <refspec>+refs/tags/*:refs/remotes/origin/tags/*</refspec>
+        <url>https://github.com/<%= @org %>/<%= @repo %></url>
```
armiller: I don't see these values are being passed to the template.

rettigs:
Rebased, looking into cleaning it up now.

templates/default/cookbook-uploader.config.xml.erb
rettigs: I think you are correct.

rettigs:
The test kitchen suite works and seems to create all the jobs, but I have not yet tested whether they actually work.

jordane:
@rettigs yep, that's about where I left this.

rettigs:
@osuosl-cookbooks/chefs So it appears functional, but it doesn't yet seem to enforce synchronization between the tags and the `metadata.rb`.  I feel like it would be ideal to have the user perform one or the other, and then have the Jenkins job finish it up.  For example, the job runs when `metadata.rb` is updated and creates the tag, or the user pushes a tag and the `metadata.rb` is updated.  It could also be entirely automatic where any new commit in master will cause the job to trigger and bump the version both in the `metadata.rb` and with a tag.

My initial inclination is the automatic version as described above, but with the ability to manually edit the `metadata.rb`.  Normally, it would perform either a minor or patch version bump (not sure what would work better), but if the commit edits the version in `metadata.rb` to a non-existent one, it will not bump it again, but rather just create the tag and upload to the Chef server.

I'd also like to set up automatic pushing from Github rather than polling.  I feel like that should be possible to automate through Octokit.

armiller:
Could we have it merge and bump the metadata and tag with a command in the pr? Like we say "bump to 2.x.x" and the bot will merge the pull request, edit metadata, tag and upload?

rettigs:
@armiller Ooh, that's a cool idea.

armiller:
@rettigs i think we could use that pull request builder plugin in jenkins to achieve something similar

rettigs:
@armiller I don't see anywhere on [the plugin's page](https://wiki.jenkins-ci.org/display/JENKINS/Github+pull+request+builder+plugin) that would indicate that it has that functionality.

rettigs:
I also don't immediately see a way to pass the bump version without that plugin.  Github can send POSTs on PR updates, and Jenkins can trigger jobs based on that, but I don't see a way to use the information in the POST in the shell script section.

rettigs:
Another possible feature is automatically updating the changelog with the PR title.

jordane:
@rettigs You definitely shouldn't tag and then have it automatically update the metadata, because then the tag would be in accurate.

The main issue with automatically updating the cookbook after a PR is that a release is cut for every PR, which is not always a desired behavior.

armiller:
@rettigs @jordane I remember trevor using the irc plugin to trigger jenkins jobs. We could use this to pass in a parameters. I was thinking:

1. Have a cookbook upload job which takes parameters: cookbook_name, version. Bumps, tags and uploads to the chef server.
2. Have a bump environment job which takes parameters: cookbook_name, version, env (optional). It would default to bumping all of the envs but we could pass in a list of specific envs as a parameter.

so we could have a jenkins channel where we can trigger:
```
!jenkins build bump_cookbook name=workstation version=2.1.4
!jenkins build bump_env name=workstation version=2.1.4 env=workstation
```
I'm not sure if the syntax is correct but i think that is how it would work.

rettigs:
@jordane Yeah, metadata then tag makes more sense.  I don't see the problem with cutting a release for each merged PR though.  That's pretty often what we want anyway, and the worst that would happen is that we simply wouldn't use all of the versions that are created, right?

rettigs:
@armiller I think the cookbook bumping would be better suited to just being in the PR rather than in IRC.  As for environment bumping, I don't know if it's a good idea to make that automatic.  Don't we usually put that in a PR to make sure that the bump gets reviewed by others first?

armiller:
@rettigs do you mean having Jenkins bump the cookbook when it merges the pr or adding commits to bump the cookbook in the pr?

rettigs:
We discussed this and agree about having the cookbook version bumps triggered from the PR, and also having a separate job that can be manually triggered that will create a PR for bumping the environments.  We could also potentially make an option in the cookbook version bump trigger to also invoke the environment bump job to make the process even more streamlined for the common case of immediately wanting the new version of the cookbook in production.

rettigs:
Also a neat idea would be automatically adding the most recent changelog entries to the environment bump PR.

templates/default/cookbook-uploader.config.xml.erb
```diff
@@ -0,0 +1,68 @@
...
+    <com.cloudbees.jenkins.GitHubPushTrigger plugin="github@1.7">
+      <spec></spec>
+    </com.cloudbees.jenkins.GitHubPushTrigger>
+    <hudson.triggers.SCMTrigger>
+      <spec>H/1 * * * *</spec>
+      <ignorePostCommitHooks>false</ignorePostCommitHooks>
+    </hudson.triggers.SCMTrigger>
+  </triggers>
+  <concurrentBuild>false</concurrentBuild>
+  <customWorkspace>/var/lib/jenkins/jenkins-data/jobs/<%= @org %>/<%= @repo %></customWorkspace>
```
rettigs: @jordane What is the purpose of defining a custom workspace?  Why not just use the default one at `/var/lib/jenkins/jobs/jobname/workspace`?

jordane: We've traditionally done this on chef jobs because reasons. It's became unnecessary sometime after this pr was created.

.kitchen.yml
```diff
@@ -47,3 +47,34 @@ suites:
   - name: ftp
     run_list:
       - recipe[osl-jenkins::ftp]
+  # In order properly test Github integration, you must export the GITHUB_USER
+  # and GITHUB_TOKEN env vars in your shell before you converge this suite.
```
jordane: If you are setting the token properly, you don't need a username.

files/default/github_pr_comment_trigger.rb
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env ruby
+
+# Takes a JSON payload from a Github PR issue webhook from stdin, reads the
+# comment, and if it matches `!bump (major|minor|patch)?` (defaults to
+# `minor`), merges the PR, bumps the Chef cookbook's metadata.rb, creates a
```
jordane: Why default to `minor`? We bump patches far more often.

files/default/github_pr_comment_trigger.rb
```diff
@@ -0,0 +1,91 @@
...
+  'major' => 0,
+  'minor' => 1,
+  'patch' => 2
+}
+
+d = JSON.load(STDIN.read)
+
+# Check if we got a valid request and get the bump level
+comment = d.fetch('comment', {}).fetch('body', '')
+match = comment.match(/^#{COMMAND}( (#{LEVELS.keys.join('|')}))?$/, 2)
```
jordane: Is there a reason why you create one match group with a space and then the other match group?

files/default/github_pr_comment_trigger.rb
```diff
@@ -0,0 +1,91 @@
...
+unless pr.mergeable
+  puts 'Error: Cannot merge PR because it would create merge conflicts.'
+  exit 1
+end
+
+# Merge the PR
+pr.merge_pull_request
+git = Git.open('.')
+git.branch(GIT_BRANCH).checkout
+git.pull(git.remote('origin'), GIT_BRANCH)
```
jordane: Did you really mean to use `GIT_BRANCH` in both of the above lines?

files/default/github_pr_comment_trigger.rb
```diff
@@ -0,0 +1,91 @@
...
+# comment, and if it matches `!bump (major|minor|patch)?` (defaults to
+# `minor`), merges the PR, bumps the Chef cookbook's metadata.rb, creates a
+# version tag, and pushes up.
+
+require 'git'
+require 'json'
+require 'octokit'
+require 'pp'
+
+GIT_BRANCH = 'master'
```
jordane: For what its worth, the JSON payload should contain the branch into which the PR is to be merged.

files/default/github_pr_comment_trigger.rb
```diff
@@ -0,0 +1,91 @@
...
+::File.write(METADATA_FILE, md)
+git.add(all: true)
+git.commit("Automatic #{level}-level version bump to v#{version} by Jenkins")
+
+# Create a version tag
+git.tag("v#{version}")
+
+# Push back to Github
+git.push(git.remote('origin'), GIT_BRANCH)
+git.push(git.remote('origin'), GIT_BRANCH, tags: true)
```
jordane: Why are you pushing tags seperately?

libraries/default.rb
```diff
@@ -0,0 +1,67 @@
...
+    h['name'] == type &&
+    h['events'] == events &&
+    h['config']['content_type'] == content_type
+  end
+
+  # If the hook exists and is outdated, delete it.
+  current_hook_exists = false
+  existing_hooks.each do |h|
+    if h['url'] != url
+      github.delete_hook(repopath, h['id'])
```
jordane: What if there are other non-jenkins hooks?

recipes/cookbook_uploader.rb
```diff
@@ -0,0 +1,115 @@
...
+  recursive true
+end
+cookbook_file github_pr_comment_trigger_path do
+  source 'github_pr_comment_trigger.rb'
+  mode '0550'
+  owner node['jenkins']['master']['user']
+  group node['jenkins']['master']['group']
+end
+
+execute_shell = "echo $payload > chef exec ruby '#{github_pr_comment_trigger_path}'"
```
jordane: So the cookbook uploader job is just this command now?

.kitchen.yml
rettigs: The username+token combo is used for `git pull`/`git push`, not for the API, which just requires the token.  Just passing a username is easier than worrying about generating an ssh keypair too.

files/default/github_pr_comment_trigger.rb
rettigs: I'm assuming people will accidentally forget to specify the bump level occasionally, and I figured that accidentally doing a minor-level bump for a patch is safer than doing a patch-level bump for a feature.

Maybe not having a default at all would be better.

files/default/github_pr_comment_trigger.rb
rettigs: Because `git push --tags` does not seem to actually push any commits like `git push` does (at least with my version and config).  I just tested `git push --tags origin master` though and it pushes both commits and tags though it seems, so I'll remove the first one.

libraries/default.rb
rettigs: I can't think of an elegant way to solve this problem since there doesn't seem to be a way to define a "comment" or anything for a hook to identify it as the cookbook uploader one.  I imagined the chances of there other hooks of type `web`, content type `form`, and only pushing `issue_comment` events would be pretty low.  I suppose a more foolproof option would be storing the hook id of the hook as an attribute on the node when it's created, but that also would accumulate hooks if the jenkins node ever got deleted and recreated since the hook ids would then be lost.

recipes/cookbook_uploader.rb
rettigs: Yeah, pretty much.  Do you think that's an issue?

files/default/github_pr_comment_trigger.rb
rettigs: Good point, I hadn't really considered merging into other branches.

files/default/github_pr_comment_trigger.rb
rettigs: I'm sorry, I don't see the issue.

files/default/github_pr_comment_trigger.rb
rettigs: It was to make both `!bump` and `!bump minor` (for example) valid.  One group contains the bump level, but I only want to match the bump level if there is a space in front of it (e.g. to prevent matching `!bumpminor`).

http://refiddle.com/refiddles/56983fc675622d3ff7a50000

Lines 1 and 3 should match.  Removing the outer group makes `!bump` require a trailing space.

files/default/github_pr_comment_trigger.rb
jordane: I don't expect it to happen often with cookbooks, but we do ocasionally merge into branches like `stable/1.x` when master gets a major bump that isn't everywhere yet.

files/default/github_pr_comment_trigger.rb
jordane: Then why not just do

`/^#{COMMAND}\s?(#{LEVELS.keys.join('|')})?$/`?

You don't need three groups.

rettigs: Your regex matches `!bumpminor` and `!bump `, which I was not intending to do.

libraries/default.rb
jordane: Is there a reason you need to wipe and re-add hooks like this?

recipes/cookbook_uploader.rb
jordane: Well, it doesn't currently do any uploading...

files/default/github_pr_comment_trigger.rb
jordane: Yeah, an error makes a lot of sense here to me.

libraries/default.rb
rettigs: The hook is only wiped if the URL is changed.  I suppose it could just be edited instead.

recipes/cookbook_uploader.rb
rettigs: The github pr comment trigger script runs the knife upload command.

files/default/github_pr_comment_trigger.rb
jordane: Hah! Goes to show me for trying to write regexes.

files/default/github_pr_comment_trigger.rb
```diff
@@ -0,0 +1,109 @@
...
+if match.nil?
+  # Exit because it wasn't a valid bump request
+  exit 0
+end
+level = match[1]
+
+# Make sure the issue is a PR
+unless d['issue'].key?('pull_request')
+  puts 'Error: Cannot merge issue; can only merge PRs.'
+  exit 1
```
armiller: Ruby has a nice `abort` command which takes a string and dumps it to stderr and exits 1.

http://ruby-doc.org/core-2.3.0/Kernel.html#method-i-abort

armiller:
From the test build the jenkins author name was "Jenkins Continuous Integration Server". Can we just shorten that to Jenkins?

rettigs:
@armiller Sounds good.

TODO: Since a build is triggered on every Github comment, perhaps I should set a limit on the number of kept builds.

rettigs:
TODO: Delete the PR branches after they are merged?

rettigs:
Success has been had.  Jenkins successfully merged [this PR](https://github.com/osuosl-cookbooks/lanparty/pull/2) using [this job](https://140.211.168.217/job/cookbook-uploader-osuosl-cookbooks-lanparty/19/console), performing every step except uploading to the Chef server.

rettigs:
TODO: Make the environment bump PR in the chef-repo too.  Also remove the comment made after merging cookbook PRs since they clutter up the newsfeed, but link to the cookbook PR from the environment bump PR so that a "referenced" link shows up for easily navigating to the environment bump PR after running a `!bump` command.

rettigs:
TODO: Have a second Jenkins job to do the environment bumping PRs, and let it be triggered from either the cookbook PR job (for convenience) or from IRC (for bumping community cookbooks and such more easily).  You can specify envs to bump in manually, or leave it blank to do all of them.

rettigs:
Where I left off with that last commit: `bump_environments.rb` is using the wrong repo (it should use `chef-repo`) and is not getting passed the right variables.

rettigs:
TODO: Make sure the CHANGELOG.md is valid markdown (e.g. escape underscores).

rettigs:
Rebased off of current master to use newer version of Jenkins.

rettigs:
Rebased again, this time ignoring commit c3144a3021129c8de7530e86a8342fe4f4764cf2 until https://github.com/osuosl-cookbooks/osl-jenkins/issues/21 is resolved since it breaks convergence.

ramereth:
@rettigs there seems to be extra commits in this rebase that have already been merged. Can you please try rebasing again? Also, are there any commits that you can squash?

rettigs:
@ramereth Those are showing up since those commits happened after c3144a3021129c8de7530e86a8342fe4f4764cf2, which I currently do not have in here since I was having issues with it.  I'll look into squashing some stuff once the PR is ready in case I need to reference previous commits before then.

rettigs:
TODO: Use Chef to automatially make `/usr/local/bin/git` the git path that Jenkins uses.

armiller:
This can be set in the system configuration for the git plugin.

rettigs:
Yep, just didn't want to lose track of any yaks.

ramereth:
@rettigs can you please rebase including that patch and also lock the version of iptables-ng (which is the original problem)?

recipes/cookbook_uploader.rb
```diff
@@ -0,0 +1,171 @@
...
+
+# Note: The `github_user` field is used for Git pull/push over https in
+# conjuction with the token, rather than an SSH key.
+begin
+  secrets = Chef::EncryptedDataBagItem.load(
+    node['osl-jenkins']['cookbook_uploader']['databag'],
+    node['osl-jenkins']['cookbook_uploader']['secrets_item']
+  )
+# TODO: Rescue a specific exception?
+rescue
```
armiller: You should use this example https://github.com/osuosl-cookbooks/osl-rsync/blob/master/recipes/default.rb#L27

recipes/cookbook_uploader.rb
```diff
@@ -0,0 +1,171 @@
...
+          "[user]\n    name = JenkinsCI\n" \
+          "[credential]\n    helper = store"
+  mode '0664'
+  owner node['jenkins']['master']['user']
+  group node['jenkins']['master']['group']
+end
+
+# Install necessary gems
+%w(git octokit).each do |g|
+  chef_gem g
```
armiller: setting this to 
```
chef_gem g do
  compile_time false
end
```
gets rid of these warnings
```
[2016-04-08T23:46:52+00:00] WARN: chef_gem[git] chef_gem compile_time installation is deprecated
[2016-04-08T23:46:52+00:00] WARN: chef_gem[git] Please set `compile_time false` on the resource to use the new behavior.
[2016-04-08T23:46:52+00:00] WARN: chef_gem[git] or set `compile_time true` on the resource if compile_time behavior is required.

[2016-04-08T23:46:53+00:00] WARN: chef_gem[octokit] chef_gem compile_time installation is deprecated
[2016-04-08T23:46:53+00:00] WARN: chef_gem[octokit] Please set `compile_time false` on the resource to use the new behavior.
[2016-04-08T23:46:53+00:00] WARN: chef_gem[octokit] or set `compile_time true` on the resource if compile_time behavior is required.
```

recipes/cookbook_uploader.rb
```diff
@@ -0,0 +1,172 @@
...
+  end
+end
+
+org_name = node['osl-jenkins']['cookbook_uploader']['org']
+chef_repo = node['osl-jenkins']['cookbook_uploader']['chef_repo']
+
+# A build is triggered on every Github comment, but will only succeed if the
+# comment was a bump request.  This message is displayed if the comment was not
+# a bump request and the build will be marked as unstable.
+non_bump_message = 'Exiting because comment was not a bump request'
```
armiller: Since this variable is a constant it should be frozen.

libraries/default.rb
```diff
@@ -0,0 +1,68 @@
...
+def public_address
+  node.fetch('cloud', {}).fetch('public_ipv4', nil) || node['fqdn']
+end
+
+# rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+def set_up_github_push(github_token, org_name, repo_name, job_name, trigger_token)
+  if ::ObjectSpace.const_defined?('Chef')
+    # Chef exists! Lets make sure Octokit is installed.
+    chef_gem 'octokit'
+  end
```
armiller: This gem is already being installed before this function is being called so we should remove this redundancy.

libraries/default.rb
```diff
@@ -0,0 +1,68 @@
+def collect_github_repositories(github_token, org)
+  if ::ObjectSpace.const_defined?('Chef')
+    # Chef exists! Lets make sure Octokit is installed.
+    chef_gem 'octokit'
+  end
```
armiller: Same here this is being installed already on line 92 of recipes/cookbook_uploader.rb

templates/default/bump_environments.rb.erb
```diff
@@ -0,0 +1,67 @@
...
+
+# Given a chef repo name, a cookbook name, a version to bump environments to,
+# and a comma-separated list of environments passed as environment variables,
+# bumps the cookbook to the given version in each of the environments (or all
+# of them if no environments are specified) and creates a PR for it on Github.
+
+require 'git'
+require 'json'
+require 'octokit'
+require 'pp' # TODO: Remove pp after done debugging
```
armiller: Is this still needed?

templates/default/github_pr_comment_trigger.rb.erb
```diff
@@ -0,0 +1,126 @@
...
+# a comma-delimited list of environments (or a * for all of them) is appended
+# to the !bump command, e.g.
+#   !bump patch *
+#   !bump minor production
+#   !bump major dev,staging
+
+require 'git'
+require 'json'
+require 'octokit'
+require 'pp' # TODO: Remove pp after done debugging
```
armiller: Can this be removed now?

templates/default/github_pr_comment_trigger.rb.erb
```diff
@@ -0,0 +1,126 @@
...
+  $stderr.puts '<%= @non_bump_message %>'
+  exit 0
+end
+level = match[1]
+envs = match[2].strip()
+
+# Make sure the issue is a PR
+unless d['issue'].key?('pull_request')
+  $stderr.puts 'Error: Cannot merge issue; can only merge PRs.'
+  exit 1
```
armiller: the `abort` command does this but in one line.
```
abort 'Error: Cannot merge issues; can only merge PRs.'
```
http://ruby-doc.org/core-2.3.0/Kernel.html#method-i-abort

templates/default/github_pr_comment_trigger.rb.erb
```diff
@@ -0,0 +1,126 @@
...
+pr = github.pull_request(repo_path, issue_number)
+if pr.merged
+  $stderr.puts 'Error: Cannot merge PR because it has already been merged.'
+  exit 1
+end
+
+# Make sure the PR can be merged without conflicts
+unless pr.mergeable
+  $stderr.puts 'Error: Cannot merge PR because it would create merge conflicts.'
+  exit 1
```
armiller: Same as above.

templates/default/github_pr_comment_trigger.rb.erb
```diff
@@ -0,0 +1,126 @@
...
+
+# Make sure the PR isn't already merged
+github = Octokit::Client.new(access_token: '<%= @github_token %>')
+repo_name = d['repository']['name']
+repo_path = d['repository']['full_name']
+issue_number = d['issue']['number']
+pr = github.pull_request(repo_path, issue_number)
+if pr.merged
+  $stderr.puts 'Error: Cannot merge PR because it has already been merged.'
+  exit 1
```
armiller: Same as above.

recipes/default.rb
```diff
@@ -18,7 +18,7 @@
 #
 
 # jenkins 1.612 and above requires java 7
-node.default['java']['jdk_version'] = '7'
+node.default['java']['jdk_version'] = '8'
```
armiller: I think you need to rebase this branch again since this change shouldn't be here.

recipes/master.rb
```diff
@@ -28,20 +28,32 @@
...
+node.set['git']['checksum'] = '106b480e2b3ae8b02e5b6b099d7a4049f2b1128659ac81f317267d2ed134679f'
+include_recipe 'git::source'
+
+# Also create a symlink since not everywhere has /usr/local/bin in their PATH.
+link '/usr/bin/git' do
+  to '/usr/local/bin/git'
+end
+
 include_recipe 'osl-jenkins::chef_ci_cookbook_template'
 include_recipe 'python'
```
armiller: This is the old version of the recipe. Please rebase against master.

rettigs:
Rebased against master.

armiller:
There is a `attributes/default.rb` file that is getting added but there is nothing in it.

rettigs:
@armiller Removed.

libraries/default.rb
rettigs: These are library functions, so there is not any guaranteed order in which they are called or whether they are called by the cookbook uploader recipe.

libraries/default.rb
rettigs: These are library functions, so the order in which they are called or whether they are called before the cookbook uploader recipe is not guaranteed.

templates/default/bump_environments.rb.erb
rettigs: I'll keep it for now.

templates/default/github_pr_comment_trigger.rb.erb
rettigs: I'll keep it for now.

libraries/default.rb
armiller: That makes no sense. Chef is run in order so those functions are being called after line 92. Have you tried removing it and seeing if it works?

Also I think we should move the `require` statements to the top of the file since we are loading `octokit` twice when it is called by the two github functions.

libraries/default.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  github = Octokit::Client.new(access_token: github_token)
+  github.auto_paginate = true
+
+  github.org_repos(org).map(&:name)
+end
+
+def public_address
+  node.fetch('cloud', {}).fetch('public_ipv4', nil) || node['fqdn']
+end
```
armiller: I think this should be put in `libraries/helpers.rb` and these other github functions should be pushed to `libraries/github.rb`

rettigs: Sure.

rettigs:
TODO: moar docs

libraries/default.rb
rettigs: The functions in the library have the potential to be called by recipes other than the cookbook_uploader recipe.  These functions depend on Octokit, so we need to enforce that we have it where the functions are.  If requiring Octokit twice is not ideal, then we should either remove the Octokit installation from the cookbook_uploader recipe or move the functions to the cookbook_uploader recipe.

armiller: Ok. So what makes the most sense to me would be moving the github functions to its own `lib/github.rb` and then installing the chef_gem first, then requiring 'octokit'. Just make sure the chef_gem is set with complie_time true so its installed before hand.

rettigs: Yep, that's what I was thinking.  As long as the library is not called `default.rb`, then Octokit won't be installed, right?

rettigs:
@armiller So one thing I wanted to clarify was the uploading of cookbooks.  Should we be using `berks upload` rather than `knife cookbook upload`?  If so, then should we be running a `berks install; berks update` as part of the process before uploading so we make sure we're uploading all the proper dependencies?

rettigs:
Docs: https://github.com/osuosl/docs/pull/262

libraries/default.rb
armiller: I see what you mean. In that case we don't want to install gems for nodes using the client recipe. We can keep the requires in the functions but it would still be good to move them the `lib/github.rb` and such. For chef_gem just make sure its getting called before those functions and everything should be fine.

armiller:
@rettigs I was thinking about that and I don't really have a good answer at the moment. I'm weary of using berks upload since it will always fetch the latests of every cookbook and will add a lot of unnecessary cookbooks to the chef-server. This may be fixed in the future by pointing berkshelf itself to our own chef server but then we would have to manually upload all of the depends and freeze them for a new cookbook.

Here is my initial thought. We move berkshelf to use source :chef_server so that way when testing we will see what depends we need to upload to the chef server. We can make another jenkins job which can upload and freeze arbitrary public cookbooks. Like in irc or something !cookbook-upload apache2 3.1.0. Then we can use the other job to create the pr to freeze them in envs.

armiller:
@rettigs so @jldugger brought up a good point regarding auth for bumping in prs. Mainly for public cookbooks where someone from the outside could trigger a build. We would need to query github for a list of users and fail if the user who triggered the comment is not in that list. We could even limit it to just the chef group so it would have to get review before being bumped.

rettigs:
@armiller Regarding creating another job, I'm not sure I understand what you mean.  Why is a second job to upload and freeze cookbooks necessary?

Also, is uploading the latest version of each cookbook a problem?  Most cookbooks have versions pinned in the production environment anyway.

rettigs:
- [ ] TODO: Don't let the cookbook uploader job trigger the environment bumper job if no environments are specified in the `!bump` command.

rettigs:
So I think the current plan is this:

Keep Berkshelf on the workstations and the cookbook uploader job to use the official supermarket, since that allows us to use newer versions of dependencies.  Then once we automate kitchen tests through Jenkins on PRs, have Berkshelf in there use our Chef server as the cookbook source.

armiller:
Here is the knife plugin: https://github.com/mdxp/knife-cleanup

rettigs:
As for uploading new versions of non-OSL cookbooks directly (rather than having Berkshelf upload them as a dependency due to a `berks upload` of a different cookbook), we could move all cookbook uploads to a different job that just uploads every cookbook version that is pinned in an environment but not on the Chef server.  However, this might be slow, and also doesn't account for dependencies.  Speaking of...

- [ ] TODO: We also need to figure out how to get the environment bumper job to automatically bump multiple cookbooks in the case where a new version of a cookbook also requires new versions of dependencies.

rettigs:
TODO: Properly handle comment events other than creating, e.g. deleting and editing.

rettigs:
We can now only trigger the job based on a list of allowed usernames, orgs, and teams.  For now, I'm planning on using `node['osl-jenkins']['cookbook_uploader']['authorized_teams'] = ['osuosl-cookbooks/staff'`.

ramereth:
@rettigs before this gets merged, I'd like to have a quick meeting about this which includes @jldugger and @armiller to talk about it in more detail. Mind setting up that meeting for me? Thanks-

templates/default/bump_environments.rb.erb
```diff
@@ -0,0 +1,66 @@
...
+  end
+  File.write(f, JSON.pretty_generate(data) + "\n")
+end
+
+# Commit and push our changes
+git.add(all: true)
+begin
+  git.commit("Automatic version bump to v#{version} by Jenkins")
+rescue Git::GitExecuteError
+  $stdout.puts 'Unable to commit; were any changes actually made?'
```
armiller: `puts` defaults to stdout. Do we want to fail here?

recipes/cookbook_uploader.rb
```diff
@@ -0,0 +1,177 @@
...
+non_bump_message = 'Exiting because comment was not a bump request'.freeze
+
+# Note: The `github_user` field is used for Git pull/push over https in
+# conjuction with the token, rather than an SSH key.
+begin
+  secrets = Chef::EncryptedDataBagItem.load(
+    node['osl-jenkins']['cookbook_uploader']['databag'],
+    node['osl-jenkins']['cookbook_uploader']['secrets_item']
+  )
+rescue
```
armiller: I still don't like that we are rescuing everything. This bypasses all data bag checks for syntax and such.

libraries/github.rb
```diff
@@ -0,0 +1,65 @@
+def collect_github_repositories(github_token, org)
+  if ::ObjectSpace.const_defined?('Chef')
+    # Chef exists! Lets make sure Octokit is installed.
+    chef_gem 'octokit'
```
armiller: Please specify compile time.

libraries/github.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  github.org_repos(org).map(&:name)
+end
+
+# rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+def set_up_github_push(github_token, org_name, repo_name, job_name,
+                       trigger_token)
+  if ::ObjectSpace.const_defined?('Chef')
+    # Chef exists! Lets make sure Octokit is installed.
+    chef_gem 'octokit'
```
armiller: same here.

libraries/github.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  require 'octokit'
+  require 'pp'
```
armiller: This require is not needed.

templates/default/github_pr_comment_trigger.rb.erb
```diff
@@ -0,0 +1,166 @@
...
+# Close the PR
+message = "Jenkins has merged this PR into `#{git_branch}` and has " \
+  "automatically performed a #{level}-level version bump to v#{version}.  " \
+  'Have a nice day!'
+github.add_comment(repo_path, issue_number, message)
+
+# Return some environment variables for the Chef environment bumper job to use.
+# If we got no envs, the file won't be created and the job won't be triggered.
+unless envs.nil?
+  File.write('envvars',
```
armiller: The previous `File.write` command was namespaced with `::File`. Is there a reason this one is not?

templates/default/bump_environments.rb.erb
rettigs: Meant to use stderr, whoops.

recipes/cookbook_uploader.rb
rettigs: So we should be providing different errors depending on whether the databag doesn't exist, or has bad syntax, etc.?  I don't see exactly what needs to be done here.  We have to capture every exception or the cookbook basically has to exit since those variables are needed everywhere.

armiller: We should only be rescuing if the chef server returns a 404 meaning the data bag does not exist. Everything else we should still raise and exception like how it is here: https://github.com/osuosl-cookbooks/osl-rsync/blob/master/recipes/default.rb#L27
Otherwise if someone pushed a bag data bag config it would silently reset the data to an empty array and we would have no way of knowing that happened.

rettigs:
Results of meeting with @ramereth, @jldugger, and @armiller:

- Let's just not automate dependencies at all.  We don't need to update dependencies often, so we can do that manually for now.
- Add a "default" set of environments for the environment bumper so if you do `!bump patch default`, 'default' will expand to a list of common environments, defined in an attribute.  This is because `*` often gets environments we don't want, like in the case of the `osl_apache` environment.
- Add some more docs to `libraries/github.rb`.

rettigs:
- [ ] TODO: Link to the environment bump PR from the merged cookbook PR.

ramereth:
@rettigs where are the tests for this suite? There seem to be zero. :(

rettigs:
@ramereth There are currently no serverspec tests.  I could probably make a couple really basic ones, but most of the functionality can't be tested without stubbing out GitHub, actually running the Jenkins's jobs, etc.

rettigs:
- [ ] TODO: Fix minor and major bumps not resetting lower numbers to 0.

ramereth:
@rettigs yes I agree however you could also add rspec tests for the actual ruby stuff which doesn't require that interaction

rettigs:
@ramereth Do you mean writing rspec for the 2 ruby scripts, `github_pr_comment_trigger` and `bump_environments`?  Can Rspec run on `.erb` files?  We need to pass in the template variables, set env vars, and also provide stdin to test those scripts, and the first one interacts heavily with GitHub (the second one only slightly).

rettigs:
- [ ] TODO: Don't fail if nonexistent environments are given to the environment bumper job?

templates/default/bump_environments.rb.erb
```diff
@@ -0,0 +1,78 @@
...
+end
+
+# Create a new branch
+Git.configure do |config|
+  config.binary_path = '<%= @git_path %>'
+end
+git = Git.open('.')
+git.branch('master').checkout
+chef_envs_string = all_envs ? 'all-envs' : chef_envs.join(',')
+git_branch = "jenkins/bump-#{cookbook}-to-#{version}-in-#{chef_envs_string}"
```
armiller: Can we short this branch name? It is crazy long especially when we use the default environments which is going to be 95% of the time.

rettigs: How do you suggest shortening it while keeping it unique? Using a hash of the environment list?

ramereth: @rettigs I think a short hash for the env list is a good idea. We can also add the list of environments as a comment in the PR it creates.

rettigs:
Whoops!

templates/default/bump_environments.rb.erb
rettigs: Done.

templates/default/bump_environments.rb.erb
```diff
@@ -44,8 +51,16 @@ Git.configure do |config|
...
-git_branch = "jenkins/bump-#{cookbook}-to-#{version}-in-#{chef_envs_string}"
+branch_hash = if is_all_envs
+                'in-all-envs'
+              elsif is_default_envs
+                'in-default-envs'
+              else
+                ''
+              end
+# Add a 10-digit hash of the version and envs so the branch names are unique.
+branch_hash += "-#{(chef_envs.to_a << version).join(',').hash % 10_000_000_000}"
```
armiller: If you are creating a hash why not just grab the first 10 characters of the hash? why the mod operation?

ramereth: +1 to what @armiller said

templates/default/bump_environments.rb.erb
```diff
@@ -0,0 +1,101 @@
...
+git_branch = "jenkins/bump-#{cookbook}-to-#{version}-#{branch_hash}"
```
armiller: I think this should be shortened to `jenkins/#{cookbook}-#{version}-#{branch_hash}`

ramereth: +1

templates/default/bump_environments.rb.erb
rettigs:
```
1_000_000.times do
  "#{47589468927645687 % 10_000_000_000}"
end

real	0m0.497s
```
```
1_000_000.times do
  "#{47589468927645687.to_s.slice(0, 10)}"
end

real	0m0.612s
```

Modding seems to be faster, and the code is about the same in length and complexity, so I vote to keep it how it is.  I'll make it a little shorter though, sure.

armiller:
@osuosl-cookbooks/chefs because this pr is getting large and difficult to review I think we should merge this into a develop branch and create individual prs for specific features.

rettigs:
@armiller My hope was that this would be merged pretty soon.  Breaking this PR up into multiple PRs would be a relatively large amount of effort for something that shouldn't be around for very long anyway.

armiller:
@rettigs I was mostly thinking about a pr for test. I'm fine with this being merged

rettigs:
@armiller In that case, I wholeheartedly agree with you.

rettigs:
@osuosl-cookbooks/chefs Thoughts on merging this PR now?  That way we get the benefits of the automation now and we have time to start deploying it before I leave.  Then we can work on the tests in a separate PR that should be nicer to load and review.

armiller:
:+1:

.kitchen.yml
```diff
@@ -54,3 +54,70 @@ suites:
...
+  - name: cookbook-uploader
+    encrypted_data_bag_secret_key_path: test/integration/encrypted_data_bag_secret
+    run_list:
+      - recipe[osl-jenkins::master]
+      - recipe[jenkins::java]
+      - recipe[jenkins::master]
+      - recipe[certificate::manage_by_attributes]
+      - recipe[osl-jenkins::haproxy]
+      - recipe[osl-haproxy::default]
+      - recipe[osl-jenkins::cookbook_uploader]
```
ramereth: Are all of these actually deps for the cookbook? If so why can't they be included in the recipe itself instead of the runlist here?

armiller: They are actually included in the recipe. I think this was added before I moved the recipes from the role to the `master` recipe.

rettigs:
One thing, I was wondering where we should be putting the attributes that we want for when this goes into production.  The attributes are currently set via the `.kitchen.yml`, but that obviously doesn't work in production.  The credentials all go in an encrypted databag, which we also need a PR for first btw, but we need a place for the rest.  Options include:

- In another recipe
- In a role
- In the production environment
- On the node itself
- Also in the databag, which requires yet more edits to this PR.

rettigs:
Talking with @armiller, we're thinking right now that we could just create another recipe called `cookbook_uploader_osl` or something that just sets some attributes and then includes `cookbook_uploader`, and that's what should go in the run list of our Jenkins server.  As for the LDAP user, it's pretty much a necessity since it seems that [multiple security realm support does not exist](https://issues.jenkins-ci.org/browse/JENKINS-15063), which would allow us to create a local user automatically.

@ramereth Thoughts?

recipes/master.rb
```diff
@@ -42,7 +42,20 @@
...
+# the Jenkins git plugin recommends 1.8.x. So we use 1.8.5.5, the latest 1.8.x.
+node.set['git']['version'] = '1.8.5.5'
+node.set['git']['checksum'] = '106b480e2b3ae8b02e5b6b099d7a4049' \
+                              'f2b1128659ac81f317267d2ed134679f'
+include_recipe 'git::source'
+
+# Also create a symlink since not everywhere has /usr/local/bin in their PATH.
+link '/usr/bin/git' do
+  to '/usr/local/bin/git'
+end
```
ramereth: Let's wrap this code block around an if statement and only do the source install on C6, but do the package install on C7 and above. That way when we do get around to upgrading the server to C7 we don't still install git by source.

recipes/master.rb
```diff
@@ -43,17 +43,24 @@
...
+                                  'f2b1128659ac81f317267d2ed134679f'
+    include_recipe 'git::source'
+
+    # Also create a symlink since not everywhere has /usr/local/bin in their PATH.
+    link '/usr/bin/git' do
+      to '/usr/local/bin/git'
+    end
+  else
+    # For CentOS 7 and up, the packaged version is new enough.
+    include_recipe 'git::default'
```
armiller: `git` is the same as `git::default`

armiller:
@rettigs, @ramereth and I talked about where to put the attributes and we thought it can just go in the cookbook_uploader recipe.

recipes/master.rb
rettigs: Done.

recipes/master.rb
rettigs: Changed to `git::package` to be more explicit and less redundant.

recipes/osl_cookbook_uploader.rb
```diff
@@ -0,0 +1,35 @@
...
+node.set['osl-jenkins']['cookbook_uploader']['chef_repo'] = 'osuosl/chef-repo'
+node.set['osl-jenkins']['cookbook_uploader']['authorized_users'] = []
+node.set['osl-jenkins']['cookbook_uploader']['authorized_orgs'] = []
+node.set['osl-jenkins']['cookbook_uploader']['authorized_teams'] = \
+  ['osuosl-cookbooks/staff']
+node.set['osl-jenkins']['cookbook_uploader']['default_environments'] = \
+  %w(production dev workstation openstack_production openstack_staging phpbb)
+
+node.set['osl-jenkins']['cookbook_uploader']['override_repos'] = \
+  %w(lanparty osl-jenkins)
```
armiller: Are you going to change this after you merge this pr?

rettigs:
@osuosl-cookbooks/chefs I went ahead and created a `jenkins` user in LDAP and [made a PR for the credentials databag](https://git.osuosl.org/osuosl-chef/data_bags/merge_requests/104).  Once that is merged, we just need to merge this PR, bump the version/environment, and add `osl_cookbook_uploader` to the run_list of our Jenkins server.  Currently, it will be deployed to just the `lanparty` and `osl-jenkins` cookbooks.  We can then add more as needed by updating the list in `osl_cookbook_uploader.rb`.

recipes/osl_cookbook_uploader.rb
rettigs: It was mentioned during meetings that we might want to roll out the deployment a few cookbooks at a time, so this can be updated by anyone after this PR is merged.  Eventually, it can be deleted so that it just does all repos in the org.

rettigs:
We need to remember to merge https://github.com/osuosl/docs/pull/262 once we merge this.

PR #4, opened 2015-04-01
Add haproxy Recipe

bramwelt:
Adds the haproxy recipe for setting the default attributes required for proxying HTTPS requests to/from Jenkins.

Includes an update to the README.


ramereth:
@bramwelt tests?

ramereth:
@bramwelt also, you might run rubocop on your recipe, I see a few violations.

bramwelt:
@ramereth I'm not sure how to test this as it is just setting attributes.

I'll fix the rubocop line violations.

recipes/haproxy.rb
```diff
@@ -0,0 +1,35 @@
...
+  'ciphers ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:' \
+  'ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS no-sslv3'
+haproxy['ssl_httpchk'] = false
+haproxy['members'] = [
+  {
+    'hostname' => 'jenkins',
+    'ipaddress' => '127.0.0.1',
+    'port' => '8080'
+  }
+]
```
armiller: This actually doesn't change any attributes. I think you would prefer doing:
```
node.default['haproxy'] = {
  'enable_ssl' => 'foo',
  'test' => 'bar'
}
```

bramwelt: That what it should be doing with the line `haproxy =
node.default['haproxy']`

Will that not actually work? Do I need to explicitly write that everytime?

armiller: the line `haproxy = node.default['haproxy']` is just copying the node attributes to a local hash. every operation afterward is just modifying the local copy. you would need to add `node.default['haproxy'] = haproxy` afterwards but it would just be easier to do:
```Ruby
node.default['haproxy'] = {

'enable_ssl' => true
'frontend_max_connections' => '2000 \n' \
  '  redirect scheme https if !{ ssl_fc }'
'ssl_incoming_port' => '443 ssl crt /etc/pki/tls/wildcard.pem ' \
  'ciphers ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:' \
  'ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS no-sslv3'
'ssl_httpchk' => false
'members' => [
  {
    'hostname' => 'jenkins',
    'ipaddress' => '127.0.0.1',
    'port' => '8080'
  }
]
}



```

bramwelt: The problem with this notation is that it will replace all the entire `haproxy` attribute hash, not update the specific attributes that need to be changed. That is definitely not what we want for a wrapper recipe.

armiller: I used to think that was the case but it turns out that attributes are deep merged so it will only update the specific attributes you list and not override the entire hash.

bramwelt: Since I have this recipe passing tests now, I'll switch over the hash and check. It could have been I had some other issue that caused me to think this wouldn't work.

bramwelt: After switching it to the hash version, I received the following error:
```
Recipe: osl-haproxy::default
  * chef_gem[chef-rewind] action install
[2015-04-09T16:16:31+00:00] INFO: Processing chef_gem[chef-rewind] action install (osl-haproxy::default line 33)
[2015-04-09T16:16:37+00:00] INFO: chef_gem[chef-rewind] installed chef-rewind at 0.0.9

    - install version 0.0.9 of package chef-rewind


  
  ================================================================================
  Recipe Compile Error
  ================================================================================
  
  Chef::Exceptions::RecipeNotFound
  --------------------------------
  could not find recipe install_ for cookbook haproxy
  

  
  Running handlers:
[2015-04-09T16:16:37+00:00] ERROR: Running exception handlers
  Running handlers complete
[2015-04-09T16:16:37+00:00] ERROR: Exception handlers complete
[2015-04-09T16:16:37+00:00] FATAL: Stacktrace dumped to /tmp/kitchen/cache/chef-stacktrace.out
  Chef Client failed. 1 resources updated in 16.168889148 seconds
[2015-04-09T16:16:37+00:00] ERROR: could not find recipe install_ for cookbook haproxy
[2015-04-09T16:16:37+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
```

As I changed nothing else, I'm confident that using that syntax won't work as expected and does not get deep merged.

bramwelt:
@osuosl-cookbooks/chefs Ready for Review

Berksfile
```diff
@@ -0,0 +1,18 @@
+source 'https://supermarket.chef.io'
+
+cookbook 'base',
+         git: 'git@github.com:osuosl-cookbooks/base'
+
```
ramereth: Why the extra line spaces?

ramereth:
@bramwelt the aliases cookbook is missing from Berksfile causing a dep problem.

Berksfile
bramwelt: This is the `rubocop` corrected way to align these lines, and I figured instead of having one of the git lines be too long, I'd make them all follow the same style.

bramwelt:
@ramereth Odd that didn't show up for me. Berks must have installed in when I was doing some previous work. Now I need to find a way to clear Berkshelf before testing...

ramereth:
@bramwelt I either run ``berks update`` or I just remove ``Berksfile.lock`` to test it fully.

Berksfile
ramereth: Its fine with the intent, but generally we haven't been making a new line space between each cookbook line after that.

bramwelt:
I will try that in the future. Thanks!

The aliases cookbook is not in the Berksfile.

Berksfile
bramwelt: Oh, that whitespace. Yeah that can totally go.

ramereth:
@bramwelt now the monitoring cookbook is missing... :)

bramwelt:
@ramereth Looks like I really needed to just copy the Berksfile from osl-haproxy, heh.

Berksfile
```diff
@@ -0,0 +1,18 @@
+source 'https://supermarket.chef.io'
+
+cookbook 'aliases',
+         git: 'git@github.com:osuosl-cookbooks/aliases'
```
ramereth: This should fit within 80char. No need to add it down with an intent. The only one that is too long I think is resource_from_hash. If you look at the osl-haproxy cookbook you'll see that format works with rubocop. Mind updating it to be more like that so we're standardized?

bramwelt: @ramereth Updated. :)

bramwelt:
@osuosl-cookbooks/chefs Calling for (hopefully) one last review.

I have rebased to master and squashed a few commits.

ramereth:
@bramwelt this LGTM but I recommend you wait to merge this until osuosl-cookbooks/osl-haproxy#14 has been merged and you've tested it against this. This changes the installation of haproxy to a package. I tested it myself locally and it's working fine. However since these are happening in tandem, it might be best to coordinate between the two. Otherwise you're going to run into trouble.

bramwelt:
I added a sanity check for the haproxy package, which passes, so I believe this is good to go. @ramereth Last two commits look good?

bramwelt:
@osuosl-cookbooks/chefs ping

jordane:
+1

ramereth:
:+1:

PR #7, opened 2015-04-10
Add X-Forwarded-Proto to haproxy config

bramwelt:
Relies on (and branches from) #4.

Adds `X-Forwarded-Proto:\ http(s)` for both http(s) frontends. This is pretty hacky, but works as expected.

Added following the advice in https://wiki.jenkins-ci.org/display/JENKINS/Jenkins+says+my+reverse+proxy+setup+is+broken

Questions
- Why is it needed in both places?
- Since all requests are redirected to https, is there a reason not to just set it under the https frontend?


armiller:
Code wise it looks good. Tests pass in kitchen. As far as the questions It may be easier to manually test on jenkins1 and seeing if you need both http and https x-forward.

bramwelt:
I'm closing as this PR has a number of issues:

* Adding the X-Forward-Proto header did not seem to fix the problem.
* This is a further reliance on a hack around the haproxy cookbook.
* HTTPs works on Jenkins and between github fine without this addition.

PR #9, opened 2015-04-30
add jenkins plugins + tests

rettigs:
Adds the build-token-root, credentials, and credentials-binding plugins, plus tests.  The first two plugins are already on jenkins1.o.o, but are not specified.

Originally was here: https://github.com/osuosl/chef-repo/pull/120


rettigs:
Ready for review.

@osuosl-cookbooks/chefs

rettigs:
The build error is caused by a Rubocop failure that is resolved by https://github.com/osuosl-cookbooks/osl-jenkins/pull/10.

bramwelt:
@rettigs This is awesome! Having more Jenkins stuff managed is always a win!

Since I definitely see us expanding this in the future to include more plugins, would you mind moving this to it's own recipe of `plugins.rb` using an attribute like `osl-jenkins:plugins` to hold them? The `master` recipe is setup only to wrap the upstream jenkins cookbook so that we get the right attributes set, and keep things out of the role.

P.S. I merged in the rubocop fixes, so if you rebase this, we can get both these changes slated for `0.6.0`

bramwelt:
@rettigs Awesome! Thanks for making those changes. The tests just need to be split out, and then this is good to go. :coffee:

rettigs:
@bramwelt What happened to the Jenkins test job?  This branch was passing yesterday, and I saw you manually ran a new build that's now failing with a weird error.

jordane:
@rettigs it is fixed now.

rettigs:
@osuosl-cookbooks/chefs Is this good to go?

jordane:
+1

fahlmant:
+1 Just a general question, is there a test besides "is a file" for this type of thing? Like an assertion that it's not empty or something that would be useful?

jordane:
@fahlmant Yes, you can look at [this](http://serverspec.org/resource_types.html#file) for documentation on how to do that.

fahlmant:
Thanks!

rettigs:
@fahlmant I suppose I could test if the file contents matches `.+` or something, but I'm not sure how useful that would be.

fahlmant:
@rettigs I was asking in a more general sense, not for this PR specifically. If it's not useful, I'm still +1 how it is

rettigs:
@fahlmant Ah, my bad.  Thanks for the review!

fahlmant:
@jordane Is this something we still need and could now merge?

jordane:
@fahlmant yes.

PR #16, opened 2015-12-30
Unfreeze version since we are running yum-cron

armiller:


armiller:
@osuosl-cookbooks/chefs we could also just update the version to what is installed now but i think it would be fine to have this updated with yum-cron. Also chef runs are broken on jenkins until we get this fixed.

rettigs:
What is the possibility of automatic `yum-cron` updates breaking things?  Does it interrupt jobs?  Does it perform major version upgrades that require configuration or plugin updates?

armiller:
I'm not sure. i just updated jenkins yesterday from a version that was about 15 releases old and did not have issues (except for it dropping support for java 1.6 but that was a one time thing.)

armiller:
@rettigs we currently have gitlab being autoupdated with yum-cron and it seems to be working just fine.

armiller:
@osuosl-cookbooks/chefs would like some more feedback.

PR #20, opened 2016-03-25
Migrate recipe and attributes from jenkins role to recipe for testing

armiller:
This also updates java to openjdk 8. I have verify it worked within the test kitchen vm.


recipes/master.rb
```diff
@@ -25,3 +25,23 @@
...
+} ]
+
+include_recipe 'java'
+include_recipe 'jenkins::master'
+include_recipe 'certificate::manage_by_attributes'
+include_recipe 'osl-jenkins::haproxy'
+include_recipe 'osl-haproxy::default'
+include_recipe 'git'
+include_recipe 'osl-jenkins::chef_ci_cookbook_template'
+include_recipe 'python'
```
ramereth: Shouldn't this be ``poise-python`` now?

.kitchen.yml
```diff
@@ -11,8 +11,15 @@ suites:
...
-      - recipe[jenkins::master]
+    attributes:
+      certificate:
+        - snakeoil:
+            cert_path: "/etc/pki/tls"
+            cert_file: wildcard.pem
+            key_file: wildcard.key
+            chain_file: wildcard-bundle.crt
+            combined_file: true
+
```
ramereth: You should remove this since you reference it in ``master.rb``.

armiller: this is needed for testing.

ramereth:
:

PR #35, opened 2016-06-19
Manually set jenkins java path to stop restarts during chef runs

armiller:
The logic in
https://github.com/chef-cookbooks/jenkins/blob/master/attributes/default.rb#L45-L53
is causing chef to restart runs every time.


armiller:
!bump patch ~

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v0.7.22.  Have a nice day!

PR #45, opened 2016-08-19
Use knife cookbook upload instead of berkshelf

armiller:
@osuosl-cookbooks/chefs


ramereth:
Without tests its difficult to say, but reading the docs this seems to be correct so :+1:

armiller:
!bump patch

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v0.7.26.  Have a nice day!

PR #48, opened 2017-02-03
Bind Serial Updater

ramereth:
Work in progress but would like some initial review on the ruby library I wrote.

Requires:
- [x] #49 
- [x] #50 

TODO
- [x] Create bumpzone recipe to create Jenkins job


ramereth:
[Maintain/DNS: Automate DNS Serial number updating](https://trello.com/c/R2QsdnGf/256-maintain-dns-automate-dns-serial-number-updating)

files/default/lib/bumpzone.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  def self.soa_oneline?(zone)
+    if regex(zone, ONELINE_REGEX)
+      true
+    elsif regex(zone, MULTILINE_REGEX)
+      false
+    else
+      false
+    end
```
jldugger: so.. in either case, this returns false? Why bother with the if?

files/default/lib/bumpzone.rb
```diff
@@ -0,0 +1,115 @@
...
+  def self.print_soa(before_serial, serial, after_serial)
+    "#{before_serial}#{serial}#{after_serial}"
+  end
+
+  def self.counter?(cur_serial)
+    counter = cur_serial.to_s.match(/\d{2}$/)
+    if counter.to_s =~ /99/
+      true
+    else
+      false
```
jldugger: This `if` block can be simplified to counter.to_s.include? '99'

files/default/lib/bumpzone.rb
```diff
@@ -0,0 +1,115 @@
...
+require 'json'
+require 'octokit'
+
+# Library to bump DNS zones
+class BumpZone
+  @commit_msg = "\n\n"
+  @warning_msg = "Warning:\n\n"
+  @warning = false
+  ONELINE_REGEX = /([\w\d\.-_]+\s+[\d]+\s+IN\s+SOA.*[\w\d\.-_]+\s+)(\d{10})(\s+[\d]+\s+[\d]+\s+[\d]+\s+[\d]+)/
+  MULTILINE_REGEX = /([\w\d\.-_]+\s+IN\s+SOA\s+[\w\d\.-_]+\s+[\w\d\.-_]+\s+\(\n\s+)(\d{10})(.*\n.*\n.*\n.*\n.*\n\s+\))/
```
jldugger: So, given that staff will eventually writing and changing zone files directly, there's a couple of problems with these regex:

 [ ] trailing whitespace after ( in an SOA record is not recognized
 [ ] hyphens in domains are not recognized (`-` is treated as a range operator within `[]` unless escaped)
 [ ] comments are not represented -- a commented out SOA record would be matched, parsed and updated

jldugger: And if you'd like a quick example of a SOA record using a hypen: `db.10.in-addr.arpa:10.in-addr.arpa.	2560	IN	SOA	ns1.auth.osuosl.org. hostmaster.10.in-addr.arpa. 1486695035 16384 2048 1048576 2560`

files/default/lib/bumpzone.rb
ramereth: If I remove this then it would return nil. I suppose that's OK in this case.

files/default/lib/bumpzone.rb
ramereth: @jldugger great catch on those. I'm working on fixing that now. Are there any other situations you can think of? I removed underscores as I realized they're not allowed in RFC.

ramereth:
@jldugger I believe I have fixed the regex

files/default/lib/bumpzone.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  def self.regex(file, regex)
+    !(file !~ regex)
+  end
+
+  def self.soa_oneline?(zone)
+    if regex(zone, ONELINE_REGEX)
+      true
+    elsif regex(zone, MULTILINE_REGEX)
+      false
```
jldugger: I'm still not sure why the body of this function isn't `regex(zone, ONELINE_REGEX) != nil`.

ramereth: What do you mean?

files/default/lib/bumpzone.rb
```diff
@@ -0,0 +1,112 @@
...
+    else
+      BumpZone.add_warning_msg("#{zone_file} does not contain an SOA\n")
+      return nil
+    end
+    new_serial = Time.new.strftime('%Y%m%d00').to_i
+    if BumpZone.max_count?(cur_serial)
+      BumpZone.add_warning_msg("#{zone_file} has a max counter of #{Time.new.strftime('%Y%m%d99')}\n")
+      return nil
+    end
+    new_serial = 1 + cur_serial.to_i if new_serial <= cur_serial.to_i
```
jldugger: This can probably be re-written as new_serial = [cur_serial.to_i + 1, new_serial].max

files/default/lib/bumpzone.rb
ramereth: That would work. I currently use the nil return value which happens if neither is found in spec tests, but I don't see any logic in the code itself that uses it.

ramereth: I think what might make more sense is to change this function to just return which regex to use instead of true/false. Thoughts?

ramereth: @jldugger ping

jldugger: Either way is fine IMO. If you change the purpose, you may wish to rename it, and if you change the return type you'd want to remove the question mark. But it's probably best not to spend too much time picking colors on this particular bikeshed.

files/default/lib/bumpzone.rb
```diff
@@ -0,0 +1,112 @@
...
+    end
+  end
+
+  def self.pr_merged(json)
+    if json['action'] == 'closed' && json['pull_request']['merged'] == true
+      nil
+    else
+      puts 'Not a merged PR, skipping...'
+      exit 0
+    end
```
Kennric: Somewhat more compact:
```
if json['action'] != 'closed' && json['pull_request']['merged'] != true
    puts 'Not a merged PR, skipping...'
    exit 0
end
```

files/default/lib/bumpzone.rb
```diff
@@ -0,0 +1,112 @@
...
+    issue_number = json['number']
+    github.pull_request_files(repo_path, issue_number)
+  end
+
+  def self.git_push
+    # Set up the git gem
+    git = Git.open('.')
+    # Commit changes
+    git.add(all: true)
+    git.commit("Automatically bumping zone serials#{BumpZone.commit_msg}")
```
Kennric: Do you want a space in here maybe? 
'"Automatically bumping zone serials #{BumpZone.commit_msg}")'

files/default/lib/bumpzone.rb
```diff
@@ -0,0 +1,112 @@
...
+  ONELINE_REGEX = /(^[\w\d\.\-]+\s+[\d]+\s+IN\s+SOA.*[\w\d\.\-]+\s+)(\d{10})(\s+[\d]+\s+[\d]+\s+[\d]+\s+[\d]+)/
+  MULTILINE_REGEX =
+    /(^[\w\d\.\-]+\s+IN\s+SOA\s+[\w\d\.\-]+\s+[\w\d\.\-]+\s+\(\s?\n\s+)(\d{10})(.*\n.*\n.*\n.*\n.*\n\s+\))/
+
+  class << self
+    attr_reader :commit_msg
+  end
+
+  def self.add_commit_msg(msg)
+    @commit_msg << msg
```
Kennric: Is formatting in these messages important? Mashing strings together might not be the most readable, but I assume these are unlikely to be read?

files/default/lib/bumpzone.rb
jldugger: I think you caught everything.

files/default/lib/bumpzone.rb
ramereth: I format them so that they are readable. What would you suggest I do differently?

files/default/lib/bumpzone.rb
ramereth: No because the commit message includes two newlines. We probably don't want to leave a blank space at the end of the top commit message.

spec/fixtures/multiline-soa
```diff
@@ -0,0 +1,7 @@
+osuosl.org      IN SOA  ns1.auth.osuosl.org. hostmaster.osuosl.org. (
```
knightsamar: Can we have 2 more versions of this ?

* One where there is a comment character after the opening parentheses and followed by comment.
* One with hyphen in the ns server name.

I believe I and @jldugger discussed this. If this has been already eliminiated as a good testcase, please ignore.

ramereth: @knightsamar can you provide an example of a test case you'd like to be included?

knightsamar: 1)there is a comment character after the opening parentheses and followed by comment.
> +osuosl.org      IN SOA  ns1.auth.osuosl.org. hostmaster.osuosl.org. ( ;here lies a comment 

2)One with hyphen in the ns server name.
> +osuosl.org      IN SOA  an-ns-with-hyphens1.auth.osuosl.org. hostmaster-with-hyphens.osuosl.org. (

files/default/lib/bumpzone.rb
```diff
@@ -0,0 +1,110 @@
...
+
+# Library to bump DNS zones
+class BumpZone
+  @commit_msg = "\n\n"
+  @warning_msg = "Warning:\n\n"
+  @warning = false
+  ONELINE_REGEX = /(^[\w\d\.\-]+\s+[\d]+\s+IN\s+SOA.*[\w\d\.\-]+\s+)(\d{10})(\s+[\d]+\s+[\d]+\s+[\d]+\s+[\d].*)/
+  MULTILINE_REGEX =
+    /(^[\w\d\.\-]+\s+IN\s+SOA\s+[\w\d\.\-]+\s+[\w\d\.\-]+\s+\((?:\s*;(?:[\s\w\d]*)?)?(?:\s*)?\n\s+)(\d{10})\
+(.*\n.*\n.*\n.*\n.*\n\s+\))/
```
ramereth: My latest update to this produces a new warning when run: ``warning: regular expression has redundant nested repeat operator '*'``

I can't figure out what's causing this, can someone please take a look and let me know? Feel free to use this [site](http://regexr.com/3fcmg) to troubleshoot and ensure it works properly too.

spec/fixtures/multiline-soa
ramereth: @knightsamar I believe I have updated the regex to handle this.

files/default/lib/bumpzone.rb
Kennric: Ah, ok, I missed the newlines - I generally don't like to trust component strings to be formatted when I construct a readable text. It looks like the component strings have the appropriate spacing/newlines, so I retract this and my previous nitpick.

spec/fixtures/multiline-soa
knightsamar: If we can make only the one with hyphens in the ns server name pass, it would be good.

Else we file these both test case as issues and keep them alive and also document them somewhere in the recipe just in case we come back to the recipe in a traceback.

ramereth: I have both use cases working now FWIW.

ramereth:
@osuosl-cookbooks/chefs if there are no other objections, I'm going to go ahead and merge this in later today.

ramereth:
!bump minor *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a minor-level version bump to v0.10.0.  Have a nice day!

PR #51, opened 2017-02-17
Switch to using Jenkins API token instead of password

ramereth:
I neglected to notice that we had switched away from using the password to using
an API token in 5a6b4e051dcfd8e6684ae49a3020cdeca15cdf4f. This resolves that
issue.


ramereth:
!bump patch *

PR #56, opened 2017-03-14
Upgrade to jenkins-2.46.1

### LINKED ISSUE #55: Update jenkins to latest LTS 2.x version
While working on development for the bumpzone jenkins jobs, I've run into several problems with the current version of jenkins we have locked (1.654-1.1). The current stable is 2.32.3 and would bring in a lot more newer features and bug fixes. The upgrade in place on a TK VM seems to be safe but we would need to look into it further.

What I'd like to know is:
- Can we move forward with updating to the latest stable?
- We will likely need to update several plugins we use
- What are the concerns if we do?

IMO this is a blocker for getting our DNS updates done. I keep running into plugins that require a newer version of jenkins.

ramereth:
This is a major upgrade from 1.x to 2.x but shouldn't cause too many issues on
current systems as long as you run any updates/fixes it needs. This is dependant
on osuosl-cookbooks/osl-munin#19 in order to use the jenkins-5.0.0 cookbook. The
newer cookbook is required to add support for Jenkins 2.x. This will also resolve #55.

New fixes:

- Added a ChefSpec test to ensure the right version of jenkins is being
  installed
- Added two ServerSpec curl tests to ensure http redirects properly and jenkins
  is running and showing the correct version
- Added the missing spec_helper require on the master suite for ServerSpec
- Fixes for ssh credentials in Jenkins 2.x
- Switch to using jenkins_command[safe-restart] instead of service to make the
  restart work more smoothly
- Improve ChefSpec tests to test for notifications

Edit: The plugin upgrades are required for Jenkins 2.x compatibility.


ramereth:
[Jenkins scripts and logic for automatically updating/checking the zonefiles](https://trello.com/c/XDo20PV7/332-jenkins-scripts-and-logic-for-automatically-updating-checking-the-zonefiles)

ramereth:
@osuosl-cookbooks/chefs this is finally ready for a full review. I'll obviously plan ahead to actually merge this and deploy it since it will impact our production jenkins server.

ramereth:
[Upgrade to jenkins-2.32.3](https://trello.com/c/6rj3HtoQ/346-upgrade-to-jenkins-2-32-3)

recipes/master.rb
```diff
@@ -67,16 +67,25 @@
   end
 end
 
+jenkins_command 'safe-restart' do
+  action :nothing
```
fahlmant: This still runs the command, right?

recipes/master.rb
```diff
@@ -67,16 +67,25 @@
...
   jenkins_plugin p do
     version v
-    notifies :restart, 'service[jenkins]'
+    notifies :execute, 'jenkins_command[safe-restart]'
   end
 end
 
+jenkins_plugin 'ssh-credentials' do
+  version '1.12'
+  notifies :execute, 'jenkins_command[safe-restart]', :immediately
```
fahlmant: Why does this need an `immediately`?

recipes/master.rb
```diff
@@ -91,8 +100,9 @@
 end
 
 # Add ssh credentials into Jenkins
-secrets['ssh'].to_a.each do |_id, cred|
+secrets['ssh'].to_a.each do |id, cred|
   jenkins_private_key_credentials cred['user'] do
+    id id
```
fahlmant: This seems a little confusing in terms of readability.

spec/unit/recipes/master_spec.rb
```diff
@@ -24,15 +27,24 @@
...
       }.each do |plugin, ver|
         it do
           expect(chef_run).to install_jenkins_plugin(plugin).with(version: ver)
         end
+        it do
+          expect(chef_run.jenkins_plugin(plugin)).to notify('jenkins_command[safe-restart]')
+        end
+      end
+      it do
+        expect(chef_run.jenkins_plugin('ssh-credentials')).to notify('jenkins_command[safe-restart]').immediately
```
fahlmant: ssh-credentials is being tested twice on the notify here. Could this be collapsed into one block with an if?

recipes/master.rb
ramereth: When notified, yes.

recipes/master.rb
ramereth: Because we need this installed before we can use ``jenkins_private_key_credentials`` further in the cookbook.

recipes/master.rb
ramereth: perhaps ``ssh_id`` ?

spec/unit/recipes/master_spec.rb
ramereth: This one specifically checks to ensure it's run immediately. The previous block is primarily checking to ensure it's being installed with the correct version. I agree that its technically somewhat duplicated but this seemed like the cleanest approach.

recipes/master.rb
```diff
@@ -67,32 +67,42 @@
...
 
 # This is necessary in order to create Jenkins jobs with security enabled.
 node.run_state[:jenkins_private_key] = secrets['jenkins_private_key'] # ~FC001
 
 # Add git credentials into Jenkins
-secrets['git'].to_a.each do |id, cred|
+secrets['git'].to_a.each do |ssh_id, cred|
   jenkins_password_credentials cred['user'] do
-    id id
+    id ssh_id
```
fahlmant: I think this should be `git_id` and the one below should be `ssh_id`

metadata.rb
```diff
@@ -10,9 +10,9 @@
 
 depends          'base'
 depends          'build-essential'
-depends          'git', '~> 4.3'
+depends          'git'
 depends          'java'
-depends          'jenkins', '~> 2.4.0'
+depends          'jenkins', '~> 5.0.0'
 depends          'osl-haproxy'
 depends          'ssh-keys'
```
knightsamar: I have always wondered why don't we lock down all versions of the cookbooks which are from upstream? If they make a change which is undesirable to our infra, we might come to know if only once things start crashing.

I understand this probably delves a bit into how we do DevOps and would like to understand how we aim to manage such a risk.

jldugger: We pin down versions in environments, IIRC. I'm not sure if the cookbook uploader automatically detects situations where dependences are updated or not, but the uploder's code knows.

ramereth: @knightsamar I agree we should be pinning at least minor versions which is what this is doing. I've been mainly adding it when I can and where it makes sense.

@jldugger the uploader won't upload any deps automatically. With the new checks I added to the chef-repo jenkins job, that should catch any dep issues we may run into.

ramereth:
!bump major *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a major-level version bump to v1.0.0.  Have a nice day!

PR #58, opened 2017-04-12
Update pinned versions of various jenkins plugins

ramereth:
This matches a working config in production. Without this we run into
downgrading problems. I will work on a separate PR to unify plugin installation
in a saner fashion so we don't have the versions all over the place.

### DIFF
--- a/recipes/bumpzone.rb
+++ b/recipes/bumpzone.rb
@@ -51,9 +51,9 @@ end
 
 {
   'mailer' => '1.20',
-  'token-macro' => '2.0',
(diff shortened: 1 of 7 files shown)


knightsamar:
So this fix actually fixes the linter which is failing on this PR and you have tested it, right?

ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.0.1.  Have a nice day!

PR #61, opened 2017-04-12
Update list of default environments the bump ~ command uses

ramereth:
Removing the old openstack environments as we don't want to bump anything there
anymore and replacing them with the openstack_mitaka. Adding testing and
phase_out_nginx since those should be safe to bump (usually).


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.0.3.  Have a nice day!

PR #64, opened 2017-05-05
Set canRoam to false so assigned node works properly

ramereth:
Otherwise this tries to run on any node and that's not intended.

### DIFF
--- a/templates/default/bumpzone.config.xml.erb
+++ b/templates/default/bumpzone.config.xml.erb
@@ -41,7 +41,7 @@
     </extensions>
   </scm>
   <assignedNode>master</assignedNode>
-  <canRoam>true</canRoam>
+  <canRoam>false</canRoam>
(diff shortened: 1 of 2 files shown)


knightsamar:
So if it fails to run on a node, it just gives up. I thought it would try running again on the actual intended node after some time?

ramereth:
@knightsamar When you select a node via the interface this additional XML field gets added. However I missed that when I added the assignedNode before. If you look at the other Jenkins job templates in the repo, you'll notice the same thing.

knightsamar:
If this value can be set in a default template, I think it would be apt to do it there too.

ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.1.4.  Have a nice day!

PR #67, opened 2017-05-23
Kennric/pr build jobs

Kennric:
This recipe will generate a jenkins job for each of a list of static sites to build. The pr_comment job is set as a downstream job, so that should be merged first.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -174,12 +174,44 @@ suites:
           override_repos:
             - test-cookbook
           github_insecure_hook: true
-          credentials:
-            github_user: <%= ENV['GITHUB_USER'] %>
-            github_token: <%= ENV['GITHUB_TOKEN'] %>
-            trigger_token: <%= ENV['TRIGGER_TOKEN'] %>
-            jenkins_user: <%= ENV['JENKINS_USER'] %>
-            jenkins_pass: <%= ENV['JENKINS_PASS'] %>
+        credentials:
+          git:
+            github_comment:
+              user: <%= ENV['GITHUB_USER'] %>
+              token: <%= ENV['GITHUB_TOKEN'] %>
+          jenkins:
+            github_comment:
+              user: <%= ENV['JENKINS_USER'] %>
+              api_token: <%= ENV['JENKINS_PASS'] %>
+              trigger_token: <%= ENV['TRIGGER_TOKEN'] %>
+      jenkins:
+        master:
+          listen_address: '0.0.0.0'
+  - name: site_pr_builder
+    run_list:
+      - recipe[osl-jenkins::site_pr_builder]
+    encrypted_data_bag_secret_key_path: test/integration/encrypted_data_bag_secret
+    provisioner: chef_zero
+    driver:
+      security_groups:
+        - default
+        - github
+    attributes:
+      osl-jenkins:
+        github_comment:
+          override_repos:
+            - test-cookbook
+          github_insecure_hook: true
+        credentials:
+          git:
+            site_pr_builder:
+              user: <%= ENV['GITHUB_USER'] %>
+              token: <%= ENV['GITHUB_TOKEN'] %>
+          jenkins:
+            site_pr_builder:
+              user: <%= ENV['JENKINS_USER'] %>
+              api_token: <%= ENV['JENKINS_PASS'] %>
+              trigger_token: <%= ENV['TRIGGER_TOKEN'] %>
       jenkins:
         master:
           listen_address: '0.0.0.0'
--- a/recipes/site_pr_builder.rb
+++ b/recipes/site_pr_builder.rb
@@ -1,0 +1,49 @@
+#
+# Cookbook:: osl-jenkins
+# Recipe:: github_comment
+#
+# Copyright:: 2017, Oregon State University
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+include_recipe 'osl-jenkins::master'
+include_recipe 'osl-jenkins::github_comment'
+
+site_pr_builder = node['osl-jenkins']['site_pr_builder']
+
+site_pr_builder['sites_to_build'].each do |site, github_url|
+  job_name = site + '_pr_builder'
+  rsync_target = "/var/www/staging.osuosl.org/#{site}-$ghprbPullId"
+  site_url = "http://#{site}-$ghprbPullId.staging.osuosl.org"
+
+  site_pr_builder_xml = ::File.join(Chef::Config[:file_cache_path], job_name, 'config.xml')
+
+  directory ::File.dirname(site_pr_builder_xml) do
+    recursive true
+  end
+
+  template site_pr_builder_xml do
+    source 'site_pr_builder_config.xml.erb'
+    mode 0440
+    variables(
+      name: site,
+      github_url: github_url,
+      site_url: site_url,
+      rsync_target: rsync_target
+    )
+  end
+
+  jenkins_job job_name do
+    config site_pr_builder_xml
+    action [:create, :enable]
+  end
+end
--- a/spec/unit/recipes/site_pr_builder_spec.rb
+++ b/spec/unit/recipes/site_pr_builder_spec.rb
@@ -1,0 +1,50 @@
+require_relative '../../spec_helper'
+
+describe 'osl-jenkins::site_pr_builder' do
+  ALL_PLATFORMS.each do |p|
+    context "#{p[:platform]} #{p[:version]}" do
+      cached(:chef_run) do
+        ChefSpec::SoloRunner.new(p) do |node|
+          node.set['osl-jenkins']['credentials']['git'] = {
+            'site_pr_builder' => {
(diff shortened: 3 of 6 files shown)


Kennric:
Looks like there are some plugin dependency/version issues still to clear up, working on those now.

Kennric:
@ramereth the test kitchen instance with this PR of jenkins complains of some outdated plugin versions. Some are general, credentials binding, etc, and some are git and github specific. We have a plugins recipe, but also some plugins being installed in the master recipe, then some recipe specific plugins in other recipes. 

Where should these plugins be added, and is there a way to test that upgrading does not break existing jobs?

The plugin version updates that make this PR happy:

```
'mailer' => '1.20',
'token-macro' => '2.1',
'github' => '1.26.2',
'build-token-root' => '1.4',
'parameterized-trigger' => '2.33',
'ghprb' => '1.37.0',
'structs' => '1.6',
'github-api' => '1.82',
'git-client' => '2.4.6',
```

ramereth:
@Kennric sigh, I was worried about this happening at some point. We created that plugin recipe but then never used it. The real fix should be a new PR that centralizes plugin versions and installation so we don't run into this problem. I was trying to avoid doing that, but unfortunately it seems as though we need to look into that now. I have a few ideas on how to go about this and if you'd like to work on it please let me know. I think that really should be done first before we fix this.

Kennric:
@ramereth rebased and updated to rely on the plugins recipe

recipes/site_pr_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+# limitations under the License.
+include_recipe 'osl-jenkins::master'
+include_recipe 'osl-jenkins::github_comment'
+
+site_pr_builder = node['osl-jenkins']['site_pr_builder']
+
+site_pr_builder['sites_to_build'].each do |site, site_url|
+  job_name = site + '_pr_builder'
+  rsync_target = "/var/www/staging.osuosl.org/#{site}-$ghprbPullId"
+  pelican_site_url = "http://#{site}-$ghprbPullId.staging.osuosl.org"
```
ramereth: Should we remove the "pelican" specific name as this also does wiki (sphinx-doc) builds too?

Kennric: I don't believe sphinx needs to be given a site_url, it's a pelican thing only (so it can link to its own theme assets). We can call it site_url instead, I suppose, just in case we have a use for it in other static sites later.

Kennric: We can change the name throughout the cookbook, but we'll need to update the pelican sites themselves before changing the env variable this job sets. I'd like to keep the env variable and pelican sites as-is until this is deployed and we are sure it is working as expected.

.kitchen.yml
```diff
@@ -183,3 +183,41 @@ suites:
       jenkins:
         master:
           listen_address: '0.0.0.0'
+      certificate:
+        - snakeoil:
+            cert_path: "/etc/pki/tls"
+            cert_file: wildcard.pem
+            key_file: wildcard.key
+            chain_file: wildcard-bundle.crt
+            combined_file: true
```
ramereth: This is no longer needed as its set globally now.

.kitchen.yml
```diff
@@ -183,3 +183,41 @@ suites:
...
+      jenkins:
+        master:
+          listen_address: '0.0.0.0'
+      certificate:
+        - snakeoil:
+            cert_path: "/etc/pki/tls"
+            cert_file: wildcard.pem
+            key_file: wildcard.key
+            chain_file: wildcard-bundle.crt
+            combined_file: true
```
ramereth: The certificate attribute is no longer needed as its set globally now.

templates/default/github-pr-comment-config.xml.erb
```diff
@@ -0,0 +1,78 @@
...
+      <bindings>
+        <org.jenkinsci.plugins.credentialsbinding.impl.UsernamePasswordMultiBinding>
+          <credentialsId>cookbook_uploader</credentialsId>
+          <usernameVariable></usernameVariable>
+          <passwordVariable>GITHUB_TOKEN</passwordVariable>
+        </org.jenkinsci.plugins.credentialsbinding.impl.UsernamePasswordMultiBinding>
+      </bindings>
+    </org.jenkinsci.plugins.credentialsbinding.impl.SecretBuildWrapper>
+  </buildWrappers>
+</project>
```
ramereth: How is this file different than ``github_comment.config.xml.erb``? When I diff it the only thing I see different is a gitlab xml block and a few changes between ``&quot;`` and ``"``.

.kitchen.yml
```diff
@@ -183,3 +183,41 @@ suites:
...
+        github_comment:
+          override_repos:
+            - test-cookbook
+          github_insecure_hook: true
+          credentials:
+            github_user: <%= ENV['GITHUB_USER'] %>
+            github_token: <%= ENV['GITHUB_TOKEN'] %>
+            trigger_token: <%= ENV['TRIGGER_TOKEN'] %>
+            jenkins_user: <%= ENV['JENKINS_USER'] %>
+            jenkins_pass: <%= ENV['JENKINS_PASS'] %>
```
ramereth: This needs to be setup similar to how I did it above in bumpzone.

templates/default/github-pr-comment-config.xml.erb
Kennric: Yes, that file shouldn't actually be included in this pr, since the comment job is merged into master already, I'll take care of that

.kitchen.yml
```diff
@@ -174,12 +174,44 @@ suites:
...
           github_insecure_hook: true
-          credentials:
-            github_user: <%= ENV['GITHUB_USER'] %>
-            github_token: <%= ENV['GITHUB_TOKEN'] %>
-            trigger_token: <%= ENV['TRIGGER_TOKEN'] %>
-            jenkins_user: <%= ENV['JENKINS_USER'] %>
-            jenkins_pass: <%= ENV['JENKINS_PASS'] %>
+        credentials:
+          git:
+            bumpzone:
```
ramereth: This needs to match the name in the template which its not currently.

.kitchen.yml
```diff
@@ -174,12 +174,44 @@ suites:
...
+        - github
+    attributes:
+      osl-jenkins:
+        github_comment:
+          override_repos:
+            - test-cookbook
+          github_insecure_hook: true
+        credentials:
+          git:
+            bumpzone:
```
ramereth: This needs to match the name in the template which its not currently.

.kitchen.yml
```diff
@@ -174,12 +174,44 @@ suites:
...
-            trigger_token: <%= ENV['TRIGGER_TOKEN'] %>
-            jenkins_user: <%= ENV['JENKINS_USER'] %>
-            jenkins_pass: <%= ENV['JENKINS_PASS'] %>
+        credentials:
+          git:
+            bumpzone:
+              user: <%= ENV['GITHUB_USER'] %>
+              token: <%= ENV['GITHUB_TOKEN'] %>
+          jenkins:
+            bumpzone:
```
ramereth: This needs to match the name in the template which its not currently.

.kitchen.yml
```diff
@@ -174,12 +174,44 @@ suites:
...
+          override_repos:
+            - test-cookbook
+          github_insecure_hook: true
+        credentials:
+          git:
+            bumpzone:
+              user: <%= ENV['GITHUB_USER'] %>
+              token: <%= ENV['GITHUB_TOKEN'] %>
+          jenkins:
+            bumpzone:
```
ramereth: This needs to match the name in the template which its not currently.

ramereth:
@Kennric can you please also add ServerSpec/ChefSpec tests?

spec/unit/recipes/site_pr_builder_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+
+      include_context 'common_stubs'
+      it 'converges successfully' do
+        expect { chef_run }.to_not raise_error
+      end
+      it 'creates the directory' do
+        %w(test-site-one_pr_builder test-site-two_pr_builder
+           test-site-three_pr_builder).each do |job|
+          expect(chef_run).to create_directory("/var/chef/cache/#{job}").with(recursive: true)
```
ramereth: Unfortunately with ChefSpec, you need to loop through these OUTSIDE of the ``it do`` block. Also, you can remove the description in the ``it do`` block as ChefSpec automatically generates that for you. The same for the tests below.

test/integration/site_pr_builder/serverspec/site_pr_builder.rb
```diff
@@ -0,0 +1,13 @@
...
+set :backend, :exec
+
+describe 'site_pr_builder' do
+  it_behaves_like 'jenkins_server'
+end
+
+describe command('curl -k https://127.0.0.1/job/site_pr_builder/ -o /dev/null -v 2>&1') do
+  its(:stdout) { should match(/X-Jenkins-Session:/) }
+  its(:exit_status) { should eq 0 }
+end
```
ramereth: You should check that each job you want to exist actually exists when you curl the site.

spec/unit/recipes/site_pr_builder_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      sites.each do |job|
+        it do
+          expect(chef_run).to create_template("/var/chef/cache/#{job}/config.xml")
+            .with(
+              source: 'site_pr_builder_config.xml.erb',
+              mode: 0440
+            )
+        end
+      end
```
ramereth: You might as well do all three of these loops as one loop which includes all three of these tests.

Kennric:
Commits squashed

!bump minor ~

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a minor-level version bump to v1.4.0.  Have a nice day!

PR #68, opened 2017-06-08
Install plugins all from one recipe

ramereth:
This pulls in all plugin installation to one place. It also replicates the
plugins that are installed on the production jenkins master. To work around the
issues described in chef-cookbooks/jenkins#534, we default to install without
plugin dependencies and directly install the exact version of every plugin we
need.

In addition:
- Remove extraneous attributes from test-kitchen
- Update git-client due to security issue


ramereth:
@Kennric you might another look. I ran into problem when I tested other suites and my new commits should fix those.

Kennric:
These changes look good to me, it might be nice to have some comments explaining what to put in each of the plugin hashes, and maybe a note to the effect that every plugin/version must be explicitly listed.

ramereth:
!bump minor *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a minor-level version bump to v1.3.0.  Have a nice day!

PR #79, opened 2017-07-06
Add jenkins_scripts for Docker cloud, OpenStack Cloud, GitHub Oauth

leel8on:
This adds a jenkins_script that configures a docker cloud. It also avoids the problem of creating duplicate entries when run multiple time which was a problem I had with the original Ansible setup. 

This Works:tm: but I wasn't sure if we wanted to pull out specific settings and make them attributes, or go about this some other way. Not sure if this something that we want to be able to reuse and should be put in another recipe and then just set the attributes in the powerci recipe.

Thoughts?

### DIFF
--- a/recipes/powerci.rb
+++ b/recipes/powerci.rb
@@ -67,9 +67,9 @@ node.default['osl-jenkins']['restart_plugins'] = %w(
   github-branch-source:2.0.5
   github-oauth:0.27
   icon-shim:2.0.3
-  matrix-auth:1.5
   authentication-tokens:1.3
-
+  embeddable-build-status:1.9
+  matrix-auth:1.5
 )
 
 node.default['osl-jenkins']['plugins'] = %w(
@@ -92,7 +92,6 @@ node.default['osl-jenkins']['plugins'] = %w(
   momentjs:1.1.1
   pipeline-build-step:2.5
   yet-another-docker-plugin:0.1.0-rc37
-  embeddable-build-status:1.9
   build-monitor-plugin:1.11+build.201701152243
   pipeline-multibranch-defaults:1.1
   pipeline-model-declarative-agent:1.1.1
@@ -276,6 +275,7 @@ jenkins_script 'Add GitHub OAuth config' do
       "hudson.model.Computer.Create",
       "hudson.model.Computer.Delete",
       "hudson.model.Computer.Disconnect",
+      "hudson.model.Computer.Provision",
       "hudson.model.Hudson.Administer",
       "hudson.model.Hudson.Read",
       "hudson.model.Item.Build",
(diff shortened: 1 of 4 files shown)


recipes/powerci.rb
```diff
@@ -88,3 +88,112 @@
 )
 
 include_recipe 'osl-jenkins::master'
+
+# Jenkins needs to be restarted before running the following script otherwise
+# the imports fail
+jenkins_command 'safe-restart'
```
ramereth: Why does it need to be restarted?

leel8on: This is the output from a converge without restarting Jenkins.

```
STDOUT: 
       STDERR: ERROR: Unexpected exception occurred while performing groovy command.
       org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
       RemoteClass: 7: unable to resolve class com.nirima.jenkins.plugins.docker.DockerTemplate
        @ line 7, column 1.
          import com.nirima.jenkins.plugins.docker.DockerTemplate
          ^
       
       RemoteClass: 9: unable to resolve class com.nirima.jenkins.plugins.docker.launcher.DockerComputerSSHLauncher
        @ line 9, column 1.
          import com.nirima.jenkins.plugins.docker.launcher.DockerComputerSSHLauncher
          ^
       
       RemoteClass: 10: unable to resolve class hudson.plugins.sshslaves.SSHConnector
        @ line 10, column 1.
          import hudson.plugins.sshslaves.SSHConnector
          ^
       
       RemoteClass: 6: unable to resolve class com.nirima.jenkins.plugins.docker.DockerCloud
        @ line 6, column 1.
          import com.nirima.jenkins.plugins.docker.DockerCloud
          ^
       
       RemoteClass: 8: unable to resolve class com.nirima.jenkins.plugins.docker.DockerTemplateBase
        @ line 8, column 1.
          import com.nirima.jenkins.plugins.docker.DockerTemplateBase
          ^
       
       5 errors
       
```
Maybe I just need to add those plugins to `['osl-jenkins']['restart_plugins']` ?

ramereth: Yup!

leel8on:
@ramereth Added the docker-plugin and its dependencies to the restart_plugins. Seems to work now without the extra restart.

leel8on:
[Jenkins/Plugin Configuration](https://trello.com/c/yQ8Q8gQy/449-jenkins-plugin-configuration)

leel8on:
@ramereth 

I'm removing the WIP label from this. There are some still things that needed to be worked on, but I think merging this in its current state would be a good idea to give @DesertBeagle a base to work off of.

ramereth:
@leel8on OK, can you please at least fix the ChefSpec errors?

leel8on:
@ramereth ChefSpec errors fixed.

ramereth:
@leel8on looks like you need to also fix ServerSpec

leel8on:
@ramereth ServerSpec now fixed.

leel8on:
Oops. I probably should have bumped instead of merged, but this recipe is not currently running anywhere and will need some modification before it is.

PR #81, opened 2017-07-12
Setting up nodes for packer pipeline

knightsamar:
This sets up nodes to run as slaves for the packer pipeline, just like the `ftp` seems to do for the ftp-related jobs.

- [x] Setup alfred user so Jenkins can connect to the nodes
- [x] Setup openstack_taster
- [x] credentials to connect use openstack_taster
- [x] packer_templates checkout
- [x] packer software
- [x] Integration tests
- [x] Unit tests

-----------------------------------
TODO: For later (in a different PR)
- [ ] Tests targeting ppc64le


knightsamar:
Compiled and released `packer v1.0.3-dev` for ppc64le at http://ftp.osuosl.org/pub/osl/openpower/packer/. We can compile only the `-dev` version due to the issue reported at https://github.com/hashicorp/packer/issues/5108

recipes/packer_pipeline.rb
```diff
@@ -37,4 +37,33 @@
...
+%w( gcc
+    gcc-c++
+  ).each do |p|
+    package p
+  end
+
+# get the openstack_taster gem (because gem_package cannot seem to use our .gem file directly
+remote_file '/tmp/openstack_taster.gem' do
+  source "https://github.com/osuosl/openstack_taster/releases/download/v#{openstack_taster_version}/openstack_taster-#{openstack_taster_version}.gem"
+end
```
ramereth: Is there any reason why we can't publish it to rubygems.org?

recipes/packer_pipeline.rb
```diff
@@ -37,4 +37,33 @@
...
+  action :install
+end
+
+# put openstack_taster in alfred's path
+execute "put openstack_taster in path" do
+  command <<-HEREDOC
+    echo 'export PATH=/opt/chef/embedded/bin:$PATH' >> ~alfred/.bashrc
+  HEREDOC
+  not_if "grep '/opt/chef/embedded/bin' ~alfred/.bashrc"
+end
```
ramereth: I don't like the idea of modifying the PATH for the alfred user. I think you should just call the binary from the absolute path and make it configurable in the jenkins job.

recipes/packer_pipeline.rb
```diff
@@ -37,4 +37,33 @@
...
 
+openstack_taster_version = node['osl-jenkins']['openstack_taster_version']
+
+# install dependencies for gem dependencies
+
+%w( gcc
+    gcc-c++
+  ).each do |p|
+    package p
+  end
```
ramereth: Instead of doing this, just include the build_essential cookbook.

recipes/packer_pipeline.rb
```diff
@@ -0,0 +1,69 @@
...
+%w( gcc
+    gcc-c++).each do |p|
+      package p
+    end
+
+# get the openstack_taster gem (because gem_package cannot seem to use our .gem file directly
+remote_file "#{Chef::Config[:file_cache_path]}/openstack_taster.gem" do
+  source "https://github.com/osuosl/openstack_taster/releases\
+/download/v#{openstack_taster_version}/openstack_taster-#{openstack_taster_version}.gem"
+end
```
ramereth: Is there any reason why we can't publish it to rubygems.org?

recipes/packer_pipeline.rb
```diff
@@ -0,0 +1,69 @@
...
+
+openstack_taster_version = node['osl-jenkins']['openstack_taster_version']
+
+# install dependencies for gem dependencies
+
+%w( gcc
+    gcc-c++).each do |p|
+      package p
+    end
```
ramereth: Instead of doing this, just include the build_essential cookbook.

recipes/packer_pipeline.rb
```diff
@@ -0,0 +1,69 @@
...
+
+# Create directory for builds and other artifacts
+directory '/home/alfred/workspace'
+
+# put the credentials for accessing openstack in alfred's home dir from the
+# encrypted databag
+node.override['osl-jenkins']['secrets_databag'] = 'osl_jenkins'
+node.override['osl-jenkins']['secrets_item'] = 'packer_pipeline_creds'
+
+arch = node['hostnamectl']['architecture'].sub!('-', '_')
```
ramereth: It's better to use ``node['kernel']['machine']`` instead of this attribute.

recipes/packer_pipeline.rb
knightsamar: We absolutely can! Will do.

ramereth: Please use a rubygems account in firfile (create one if there isn't one already).

recipes/packer_pipeline.rb
knightsamar: Would adding an alias be okay ?

ramereth: I'd rather stick with absolute paths that are configurable via the job.

attributes/packer_pipeline.rb
```diff
@@ -0,0 +1,6 @@
+default['osl-jenkins']['packer_pipeline']['openstack_taster_version'] = '0.0.2'
+default['packer']['version'] = '1.0.2'
```
ramereth: This belongs in the recipe not the attribute.

knightsamar: How do we determine what goes in the recipes and what goes in the attribute files ?

ramereth: The general rule of thumb: If it's an upstream cookbook attribute, that should be in the recipe. If it's an attribute in the same namespace as the cookbook you're working on, that can be in the attributes.

knightsamar:
@ramereth ready for second re-review.

knightsamar:
So @ramereth in my `.kitchen.yml`, I have:

```
platforms:
  - name: centos-power-7.3
    driver_plugin: openstack
    driver_config:
      image_ref: "CentOS 7.3 LE"
    transport:
      username: centos
```
and
```
  - name: packer-power
    encrypted_data_bag_secret_key_path: test/integration/encrypted_data_bag_secret
    data_bags_path: test/integration/data_bags
    run_list:
      - recipe[osl-jenkins::packer_pipeline]
    includes:
      - centos-power-7.3
```
but the packer-power test suite simply does not show up in `kitchen list`!  I have my openstack config for POWER in a different platform. What could be wrong ?
I have pushed the changes to this branch. Could you please tell me how you use two different openstack clusters?

ramereth:
@knightsamar it's likely due to our use of the operating system supports in ``metadata.rb``. If you rename the platform to ``centos-7.3`` it should show up.

knightsamar:
@ramereth so I fixed the .kitchen.yml and the instance is listed but even though I can list the image, kitchen has an error saying "Image not found". What am I missing ?

Also, have you used this somewhere in another cookbook ? I want to have both platforms and their test suites to be available and converge-able when pointed to the right openstack cluster vars.

Please ignore the merged commits from another branch. I will remove them.

knightsamar:
@ramereth so I have decided to work on the POWER specific test suite and the overall integration + unit tests, later since I need to deliver the whole pipeline by Friday and I cannot test the whole pipeline without having real slaves. I just got enough done using openstack vms as slaves.

Could you please review it as it is and let me know if it's good to merge ?

knightsamar:
If I can't merge this, I won't be able to move ahead with https://github.com/osuosl/packer-templates/pull/26 and hence the whole pipeline is clogged :)

.kitchen.yml
```diff
@@ -2,6 +2,14 @@
...
   flavor_ref: 'm1.medium'
 
+platforms:
+  - name: centos-7.3
+    driver_plugin: openstack
+    driver_config:
+      image_ref: "b0cfefd8-ce5b-4495-bdaf-baf163676819"
+    transport:
+      username: centos
+
```
ramereth: I don't like the idea of adding in a PPC64 image in here permanently.

recipes/packer_pipeline_node.rb
```diff
@@ -0,0 +1,88 @@
...
+# put the credentials for accessing openstack in alfred's home dir from the
+# encrypted databag
+node.override['osl-jenkins']['secrets_databag'] = 'osl_jenkins'
+node.override['osl-jenkins']['secrets_item'] = 'packer_pipeline_creds'
+
+include_recipe 'sbp_packer::default'
+node.override['packer']['version'] = '1.0.2'
+
+arch = node['kernel']['machine']
+if arch == 'ppc64le'
```
ramereth: I would change these two lines to just:

``` ruby
if node['kernel']['machine'] == 'ppc64le'
```

recipes/packer_pipeline_node.rb
```diff
@@ -0,0 +1,88 @@
...
+  node.override['packer']['url_base'] = node['osl-jenkins']['packer_pipeline']['packer_ppc64le']['url_base']
+  node.override['packer']['checksum'] = node['osl-jenkins']['packer_pipeline']['packer_ppc64le']['sha256sum']
```
ramereth: Why doesn't ``node.default`` work in this case? I'd like to reserve override unless we really need it.

recipes/packer_pipeline_node.rb
```diff
@@ -0,0 +1,88 @@
...
+end
+
+openstack_credentials = credential_secrets[arch]
```
ramereth: Use

``` ruby
credential_secrets[node['kernel']['machine']]
```

recipes/packer_pipeline_node.rb
```diff
@@ -0,0 +1,88 @@
...
+include_recipe 'build-essential::default'
+
+# get the openstack_taster gem (because gem_package cannot seem to use our .gem file directly
+remote_file "#{Chef::Config[:file_cache_path]}/openstack_taster.gem" do
+  source "https://github.com/osuosl/openstack_taster/releases\
+/download/v#{openstack_taster_version}/openstack_taster-#{openstack_taster_version}.gem"
+end
+
+# install openstack_taster
+gem_package 'openstack_taster' do
```
ramereth: Not chef_gem?

recipes/packer_pipeline_node.rb
```diff
@@ -0,0 +1,88 @@
...
+  source "#{Chef::Config[:file_cache_path]}/openstack_taster.gem"
+  options '--no-user-install'
+  clear_sources true
+  action :install
+end
+
+# setup qemu so that we can build images!
+if arch == 'x86_64'
+  include_recipe 'base::kvm'
```
ramereth: Use this instead:

``` ruby
include_recipe 'base::kvm' if node['kernel']['machine'] == 'x86_64'
```

recipes/packer_pipeline_node.rb
```diff
@@ -0,0 +1,88 @@
...
+elsif arch == 'ppc64le'
+  log 'qemu on ppc64le' do
+    message 'Assuming this ppc64le node has been already setup as a OpenStack compute node \
+    and skipping qemu installation'
+    level :info
```
ramereth: Remove the entire if/elseif block and just add this:

``` ruby
only_if { node['kernel']['machine'] == 'ppc64le'
```

recipes/packer_pipeline_node.rb
knightsamar: i am using the arch variable below on 51

recipes/packer_pipeline_node.rb
knightsamar: It should but I remember facing a similar problem sometime back.

Will test it out.

knightsamar:
Fixed all things as per review. Now going ahead to try and merge this in interest of time and deadlines.

knightsamar:
!bump minor *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a minor-level version bump to v1.7.0.  Have a nice day!

PR #83, opened 2017-07-24
Bring both public & vlan42 network interfaces

knightsamar:
Also allow connections from github for better testing

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -22,10 +22,12 @@ suites:
   - name: master
     run_list:
       - recipe[osl-jenkins::master]
+    driver_config:
+      user_data: userdata.txt
     driver:
       network_ref:
-        - public
         - vlan42
+        - public
       security_groups:
         - default
         - github
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -22,6 +22,13 @@ suites:
   - name: master
     run_list:
       - recipe[osl-jenkins::master]
+    driver:
+      network_ref:
+        - public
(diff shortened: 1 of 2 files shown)


knightsamar:
@doublej472 yeah I noticed that too. We have to run a `dhclient -v {interface}` to actually get the ip.

ramereth:
You're essentially running into [this problem](https://barakme.wordpress.com/2014/08/19/configuring-an-openstack-vm-with-multiple-network-cards/) and it's something I've known for a while. Although it's odd that you got an ip for eth1 but not eth0. The Cloud Foundry project ran into this as well and they just worked around it by manually running ifup on the interface.

There's a few ways we can work around this for test-kitchen:
1. Add a userdata.txt file which includes the ifup commands you need manually during booth using [cloud-init](https://cloudinit.readthedocs.io/en/latest/topics/examples.html#run-commands-on-first-boot)
2. Use the ``ifconfig`` resource in chef and bring the interface up during the converge phase using a test cookbook.

ramereth:
@knightsamar  @doublej472 I believe what I just pushed should fix the issue now. I also changed this to be a new global setting for all instances in this cookbook (which makes sense if you're testing with github).

ramereth:
@doublej472 ya, I'm not sure how to work around it but it seems the first default route gets the preference which is the important part.

doublej472:
`ip r del default via 10.1.100.1` gets rid the private default route, if you want that in the `userdata.txt` also.

PR #87, opened 2017-08-04
the real true packer pipeline job

knightsamar:
This is dependent on

1) osuosl/packer-templates#26. Once that is merged, I will change the reference to use the file from the master branch instead of the that PR's branch.
2) an MR for the authentication token.


knightsamar:
Actually this is ready to for a first cut since this allows us to test the job on production.

ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.8.3.  Have a nice day!

PR #90, opened 2017-08-10
install openstack_taster as a chef_gem explicitly

knightsamar:
This would also require installing the dependencies at compile_time
since chef installs chef_gem at compile time too.

It seems until now Chef was simply falling back to installing openstack_taster
by default as a chef_gem when a system ruby was not available.
This didn't hit our radar until we saw a machine which had both rubies.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -301,3 +301,13 @@ suites:
       - recipe[osl-jenkins::packer_pipeline_node]
     excludes:
       - centos-6.9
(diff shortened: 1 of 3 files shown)


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.8.5.  Have a nice day!

PR #92, opened 2017-08-11
Update to Jenkins 2.60.2

ramereth:
Other fixes included:

- Use yum-version-lock to ensure we don't accidentally upgrade jenkins

Something to keep in mind is that this is the first release which requires Java 8 to run according to the [Changelog](https://jenkins.io/changelog-stable/). I checked a while back and I'm pretty sure all of our clients are using Java 8 but it wouldn't hurt to double check.

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -20,6 +20,7 @@ depends          'poise-python'
 depends          'git'
 depends          'sudo'
 depends          'sbp_packer', '= 1.4.7'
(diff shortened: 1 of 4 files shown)


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.8.8.  Have a nice day!

PR #93, opened 2017-08-11
use yum-qemu-ev instead of base for setting qemu-kvm

knightsamar:


recipes/packer_pipeline_node.rb
```diff
@@ -102,7 +102,7 @@
 end
 
 # setup qemu so that we can build images!
-include_recipe 'base::kvm' if node['kernel']['machine'] == 'x86_64'
+include_recipe 'yum-qemu-ev' if node['kernel']['machine'] == 'x86_64'
```
ramereth: Two things:

- [ ] You need to keep ``base::kvm`` but move it AFTER ``yum-qemu-ev``
- [ ] Remove the architecture statement, this should work on ppc64le the same.

knightsamar: @ramereth  Done!

ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.8.9.  Have a nice day!

PR #98, opened 2017-08-25
openstack client for pipeline nodes

knightsamar:
This installs the [python-openstackclient](https://pypi.python.org/pypi/python-openstackclient) package using `poise-python`, _if_ it is not already installed.
Hence it would require us to `scl-enable` python everytime we want to use it.


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.8.11.  Have a nice day!

PR #101, opened 2017-08-31
Setup chefdk on packer-pipeline nodes

knightsamar:
This is required to provide `berks` and other utilities which are used in building packer images which require chef provisioning like the mitaka-aio and mitaka-identity images

### DIFF
--- a/recipes/packer_pipeline_node.rb
+++ b/recipes/packer_pipeline_node.rb
@@ -29,6 +29,9 @@ include_recipe 'build-essential::default'
 include_recipe 'yum-qemu-ev::default'
 include_recipe 'base::kvm'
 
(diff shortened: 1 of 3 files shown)


DesertBeagle:
Oh, I see what you mean now. I think this should be added to the run list for the node rather than included in the recipe.

knightsamar:
@DesertBeagle actually i am also a bit confused about the placement because if you see the recipe, we have been including the `base::kvm` and other such recipes *inside* it.

DesertBeagle:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.8.13.  Have a nice day!

PR #112, opened 2017-11-27
Fix ChefSpec tests

ramereth:
This adds osl-syslog to the Berksfile which was added as a dep to osl-haproxy.
In addition, don't use runit >= 4.0.0 as it breaks ChefSpec tests along with
some other cookbooks that use this cookbook (as mentioned in
osuosl-cookbooks/proj-ros#51).


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.9.6.  Have a nice day!

PR #115, opened 2017-12-18
Adjust timeouts for packer jobs

ramereth:
There's no need to set the quiet period for this job. In addition, let's
increase the amount of jobs we can run from 2 per hour to 4 per hour. I think
this is a more sensible configuration for now.


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.9.9.  Have a nice day!

PR #122, opened 2018-02-23
Bump to jenkins-2.89.4

ramereth:
Also do the following security updates on the plugins:

- credentials-binding
- junit
- scm-api
- workflow-support

### DIFF
--- a/attributes/plugins.rb
+++ b/attributes/plugins.rb
@@ -19,7 +19,7 @@ default['osl-jenkins']['plugins'] = %w(
   cloudbees-folder:6.0.3
   command-launcher:1.2
   conditional-buildstep:1.3.1
-  credentials-binding:1.13
+  credentials-binding:1.15
   cvs:2.12
   display-url-api:1.1.1
   docker-commons:1.8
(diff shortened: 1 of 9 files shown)


attributes/plugins.rb
```diff
@@ -69,7 +69,7 @@
   pipeline-utility-steps:1.4.0
   plain-credentials:1.4
   run-condition:1.0
-  scm-api:2.2.0
+  scm-api:2.2.6
```
doublej472: I don't see a security update for this plugin, but there isn't any major, breaking changes, so this is fine.

doublej472: I'm looking at https://plugins.jenkins.io/scm-api for the changes BTW

ramereth: I believe this was a dependency that was pulled in for one of the plugins that got updated.

ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.9.16.  Have a nice day!

PR #126, opened 2018-03-07
Update subversion and git plugins for security patches

ramereth:
This also included updating/adding a few other dependant plugins to get this
working properly.

### DIFF
--- a/attributes/plugins.rb
+++ b/attributes/plugins.rb
@@ -5,13 +5,15 @@
 #     println ("${plugin.getShortName()}:${plugin.getVersion()}")
 # }
 default['osl-jenkins']['restart_plugins'] = %w(
-  credentials:2.1.13
+  structs:1.14
+  credentials:2.1.16
   ssh-credentials:1.13
 )
 default['osl-jenkins']['plugins'] = %w(
   ace-editor:1.1
   ant:1.8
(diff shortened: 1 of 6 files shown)


attributes/plugins.rb
```diff
@@ -5,13 +5,15 @@
 #     println ("${plugin.getShortName()}:${plugin.getVersion()}")
 # }
 default['osl-jenkins']['restart_plugins'] = %w(
-  credentials:2.1.13
+  structs:1.14
+  credentials:2.1.16
```
doublej472: It looks like version [2.1.15](https://plugins.jenkins.io/credentials) might end up breaking things, will this need to be handled?

ramereth: I'm not sure I'm following what you're referring to here.

doublej472: This is specifically what I'm referencing:
```
NOTE: all existing credentials fingerprint tracking history will be lost.
```

ramereth: I don't think we need to worry about that.

ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.9.19.  Have a nice day!

PR #133, opened 2018-05-14
Revert "Use SSH instead of HTTPS for pulling github repos"

eldebrim:
The https site is required somewhere in the build process, so this will revert the change that made it use SSH until we find a better way to pull from private repos.


eldebrim:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.10.6.  Have a nice day!

PR #137, opened 2018-06-27
Always upgrade openstack_taster

ramereth:
Since we control when new releases of openstack_taster, we don't need to lock
the version in the cookbook. This removes the requirement of bumping this
cookbook on every new release. It also ensures that it always uses the latest
version which it currently does even if you bump the version.

Also move jenkins checking code into the shared_examples block since that was breaking tests.


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v1.10.10.  Have a nice day!

PR #145, opened 2018-08-08
Create new GPU labels for CUDA 9.1 and 9.2

ramereth:
We've run into an issue where one of our users is having problems running jobs
on two different machines in the same label [1]. This is due to the fact that
these two servers have different CUDA drivers installed. One that supports 9.2
and another that doesn't. To work around this issue we're going to create two
new labels which point to the specific hosts which should work.

I needed to refactor the code for SGE a bit to deal with adding more than one
SGE cloud, but other than that, it's the same logic as before.

[1] https://support.osuosl.org/Ticket/Display.html?id=30202

### DIFF
--- a/libraries/sge.rb
+++ b/libraries/sge.rb
@@ -1,15 +1,20 @@
 module OSLSGE
   module Helper
     # rubocop:disable Metrics/ParameterLists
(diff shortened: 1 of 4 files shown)


ramereth:
!bump patch *

PR #146, opened 2018-08-22
Update to Jenkins 2.121.3

ramereth:
This also provides updates to the following plugins:

- credentials
- ssh-credentials
- apache-httpcomponents-client-4-api
- bouncycastle-api
- ghprb
- git
- github
- github-branch-source
- ssh-agent
- workflow-step-api

And a few other dependent plugins depending on the recipe.

In addition, this provides a better fix for testing on the powerci and ibmz suites.


ramereth:
[#30194: not able to login to Jenkins](https://trello.com/c/RlBbOZh9/912-30194-not-able-to-login-to-jenkins)

attributes/plugins.rb
```diff
@@ -28,12 +28,12 @@
   docker-workflow:1.10
   durable-task:1.17
   external-monitor-job:1.4
-  ghprb:1.40.0
-  git:3.8.0
+  ghprb:1.42.0
```
eldebrim: Do we want this plugin in the array in the `ibmz` and `powerci` recipes?

ramereth: It's not needed there.

recipes/ibmz_ci.rb
```diff
@@ -68,11 +68,11 @@ class ::Chef::Recipe
   email-ext:2.62
```
eldebrim: Since we updated the `cloud-stats` plugin to 0.14 in the `powerci` recipe, should this be updated as well?

ramereth: It's not required so I'd rather not.

PR #147, opened 2018-10-15
Set SSH connection timeout to 600 seconds

ramereth:
This is to work around a leak issue we're running into on the powerci jenkins.
It seems to be related to this [1] and has a suggested workaround which is this.
We've already updated the ssh-agents plugin to the version higher than they
suggest.

[1] https://support.cloudbees.com/hc/en-us/articles/115002327291-SSH-Agents-leak-Computer-threadPoolForRemoting-threads


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v2.0.1.  Have a nice day!

PR #151, opened 2018-11-02
Remove docker-gpu-cuda91 label

ramereth:
CGRB is going to be taking the machine this label uses out of rotation so that
we can use it as an OpenStack GPU node. There are no current jobs using this
label so it should be safe to remove.


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v2.0.5.  Have a nice day!

PR #153, opened 2018-11-19
Fix bug when templates use an inline script

ramereth:
This ensures that we don't run into an issue where scripts returns nil when
using an inline script. This was happening with this job [1].

[1] https://jenkins.osuosl.org/job/packer_pipeline/579/console


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v2.0.7.  Have a nice day!

PR #161, opened 2019-04-15
Refactor plugins attribute to use hash instead of array

ramereth:
This simplifies dealing with plugins between our standard deployment vs. powerci
and ibmz-ci. This also allows us to ensure we have the same (or nearly the same)
versions of the plugins installed globally. I've been meaning to make this
change for ages!

In addition, updated ircbot to 2.30.


ramereth:
!bump minor *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a minor-level version bump to v2.2.0.  Have a nice day!

PR #164, opened 2019-04-16
Set alfred gid/uid to something static

ramereth:
This is needed for utilizing the docker-custom-build-environment plugin. We need
to have a known static UID in containers on all hosts acting as a jenkins node.
I will be deploying a one-time manual fix on each current node as this doesn't
fix any current running processes or change the gid for the home directory
correctly from the current setup.

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -22,24 +22,27 @@ node.default['java']['jdk_version'] = '8'
 
 include_recipe 'java'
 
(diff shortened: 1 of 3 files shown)


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v2.2.3.  Have a nice day!

PR #158, opened 2019-01-26
Convert cookbook_uploader to library and add tests

jerry-peng:
- [X] Convert template files to library ruby files
- [X] Change recipe so it uses the library files.
- [x] Make sure kitchen test pass
- [X] Add rspec tests for library files.

PR #169, opened 2019-05-24
Jerryp/initial cookbook uploads

jerry-peng:
Dependent on #158 
- [x] Added logic check whether cookbook is new, and if so, update environments.
- [x] Added tests to verify that the environments changes if new cookbook is uploaded.


ramereth:
@JerryPeng0112 can you please fix the cookstyle issues?

PR #175, opened 2019-09-19
Remove reliance on chef-sugar since we don't need it here anyway

ramereth:
We never added it in the metadata.rb so this is the only change that's needed.


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v2.3.1.  Have a nice day!

PR #177, opened 2019-12-02
Ensure we only install < 2.0 for openstack_taster

ramereth:
For preparing for Chef 14, we need to update the version of InSpec we depend on.
We'll be making that change in 2.0 of openstack_taster so let's ensure the
current environment doesn't accidentally pull in the newer version once it's
pushed.

NOTE: The automatic tests will likely fail since this cookbook hasn't been updated to 14 yet, however this is a pre-req for it.

### DIFF
--- a/recipes/packer_pipeline_node.rb
+++ b/recipes/packer_pipeline_node.rb
@@ -110,6 +110,7 @@ end
 # install openstack_taster
 chef_gem 'openstack_taster' do
   compile_time true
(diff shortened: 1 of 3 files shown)


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v2.3.2.  Have a nice day!

PR #181, opened 2019-12-06
Use username/password for cli executor instead of ssh key

ramereth:
We switched to using the http protocol from remoting so this needs to be
updated.


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v2.5.1.  Have a nice day!

PR #185, opened 2020-01-07
Use vhost hostname for jenkins server webhooks

ramereth:
By default, we're using the FQDN for setting the hostname used in Github
webhooks. This becomes a problem when deploying more than one OSL Jenkins server
for a migration. It makes more sense to just point to the hostname we actually
use for our Jenkins instance.

This should make it easier to do major server platform upgrades on our primary
Jenkins server.

### DIFF
--- a/attributes/cookbook_uploader.rb
+++ b/attributes/cookbook_uploader.rb
@@ -2,6 +2,9 @@
 ### Attributes that you probably want to override:
 ###
 
+# String; jenkins server vhost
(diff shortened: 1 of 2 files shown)


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v2.6.1.  Have a nice day!

ramereth:
[Upgrade jenkins host to centos7](https://trello.com/c/zQuzAf7a/1422-upgrade-jenkins-host-to-centos7)

PR #201, opened 2021-01-13
Temporarily disable installation of the openstack_taster gem

ramereth:
We are not currently using it and I'm not planning on using it in the future however I'm not ready to completely remove the code here just in case.

This should fix chef runs that are currently failing due to this.

In addition, cookstyle fixes.


ramereth:
!bump patch *

osuosl-manatee:
Jenkins has merged this PR into `master` and has automatically performed a patch-level version bump to v2.11.2.  Have a nice day!

PR #220, opened 2023-06-06
Add AlmaLinux 8

ArtofBugs:


ArtofBugs:
Also unable to run InSpec tests here, so I won't be touching this much for now.

ramereth:
@ArtofBugs is this still the case?

ArtofBugs:
> @ArtofBugs is this still the case?
This has been fixed now; I've been working on other things recently so I haven't touched this in a bit.

PR #231, opened 2024-07-09
Test Kitchen Config Refactor

zwall-bp:
All tests green for all drivers, except for
* `powerci`
* `ibmz-ci`

Both have errors that are consistent across all drivers.


ramereth:
@zwall-bp it works for me... Do you have the environment variables set that are referenced in the kitchen.yml?

PR #255, opened 2026-08-01
data-bags: build only the default branch

ramereth:
The data-bags multibranch job now only ever discovers the default branch: MR discovery is removed and branch discovery is narrowed with a `headRegexFilter('master|main')` (full-match, so a master → main rename needs no change here). Jenkins' role reduces to what it should be — re-validate and deploy on every merge — while GitLab CI, which runs the same checks, becomes the sole merge request gate.

## Deployment notes

- After converge, the job's existing MR/branch items become orphans and age out via the 7-day orphaned-item policy (or delete them at the next scan).
- Since Jenkins no longer reports any status on MRs, consider enabling **"Pipelines must succeed"** in the data_bags project's merge request settings so failing GitLab CI blocks the merge — one-time GitLab UI setting.
- Companion comment-only update in the data_bags repo (Jenkinsfile header + .gitlab-ci.yml).

## Test

- `headRegexFilter` symbol and full-match semantics verified against the scm-api plugin source
- Inspec asserts the filter is present and MR discovery is gone; chefspec 594 examples, 0 failures; cookstyle clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


osuosl-manatee:
Jenkins has merged this PR into `master` and performed a patch-level version bump to v3.8.1. Environment bump queued for: default. :key: Merged using ramereth's admin bypass of branch protection (required checks were green).

